PR #2, opened 2015-11-09
Add support for RegExps

evanlucas:


src/llv8.cc
```diff
@@ -457,6 +458,11 @@ std::string HeapObject::Inspect(bool detailed, Error& err) {
     return pre + fn.Inspect(detailed, err);
   }
 
+  if (type == v8()->types()->kJSRegExpType) {
+    JSRegExp re(this);
+    return pre + re.Inspect(detailed, err);
```
indutny: I think you need to do `"<JSRegExp>"` here instead of `re.Inspect`. Or better implement `Inspect` method! ;)

evanlucas: That is my plan. I am still trying to wrap my head around all of it. The current inspection looks like:

```
(lldb) v8 i 0x00002c0eb9123c69
0x00002c0eb9123c69:<Object: RegExp properties {
    .source=0x0000105f9260a241:<String: "blahblah">,
    .global=0x00001253415043c9:<false>,
    .ignoreCase=0x00001253415043c9:<false>,
    .multiline=0x00001253415043c9:<false>,
    .lastIndex=<Smi: 0>}>
```

It just walks the properties of the object. Before, it just showed unavailable.

indutny: Aaaah, so it calls `JSObject::Inspect`. Nice!

evanlucas: Yep, there are others that do not work like this. Want me to include those in this same PR?

indutny:
LGTM

indutny:
Landed in 2a65f509bd2f7e355c7b8740af7ddc4776c6523c, thank you!

PR #5, opened 2015-11-11
add cmd : v8 source list

bmeck:
adds `v8 source list` when on a v8 stack frame

uses `-l $lineno` to match lldb, does not support giving memory pointer to the command yet.

### DIFF
--- a/src/llnode.cc
+++ b/src/llnode.cc
@@ -113,6 +113,7 @@ bool PrintCmd::DoExecute(SBDebugger d, char** cmd,
   return true;
 }
 
+
 bool CodeMap::DoExecute(SBDebugger d, char** cmd,
                         SBCommandReturnObject& result) {
   SBTarget target = d.GetSelectedTarget();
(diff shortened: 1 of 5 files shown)


src/llnode.cc
```diff
@@ -137,6 +136,91 @@ bool CodeMap::DoExecute(SBDebugger d, char** cmd,
   return true;
 }
 
+bool ListCmd::DoExecute(SBDebugger d, char** cmd,
```
indutny: Let's have two newline separation between functions.

src/llnode.cc
```diff
@@ -120,7 +120,6 @@ bool CodeMap::DoExecute(SBDebugger d, char** cmd,
     result.SetError("No valid process, please start something\n");
     return false;
   }
-
```
indutny: Irrelevant change.

bmeck: was during rebasing

src/llnode.cc
```diff
@@ -137,6 +136,91 @@ bool CodeMap::DoExecute(SBDebugger d, char** cmd,
...
+    cmd += full_cmd;
+    interpreter.HandleCommand(cmd.c_str(), result, false);
+    return true;
+  }
+  
+  // V8 frame
+  v8::Error err;
+  v8::JSFrame v8_frame(&llv8, static_cast<int64_t>(frame.GetFP()));
+  
+  const static uint32_t DISPLAY_LINES = 4;
```
indutny: Do you mind if we will call it `kDisplayLines`? This is how we do it in all other files.

src/llv8-inl.h
```diff
@@ -87,6 +87,10 @@ inline int64_t Map::GetType(Error& err) {
 }
 
 
+inline JSFunction JSFrame::GetFunction(Error& err) {
+  return v8()->LoadValue<JSFunction>(raw() + v8()->frame()->kFunctionOffset, err);
+}
+
```
indutny: Two newlines, please.

src/llv8-inl.h
```diff
@@ -87,6 +87,10 @@ inline int64_t Map::GetType(Error& err) {
 }
 
 
+inline JSFunction JSFrame::GetFunction(Error& err) {
+  return v8()->LoadValue<JSFunction>(raw() + v8()->frame()->kFunctionOffset, err);
```
indutny: 80 column limit? What do you think about using an `ACCESSOR` macro instead?

src/llv8.cc
```diff
@@ -352,19 +402,69 @@ std::string SharedFunctionInfo::ToString(Error& err) {
...
+    else {
+      length++;
+    }
+  }
+  if (line_i >= start_line && length != 0 && lines_found < line_limit) {
+    lines[lines_found] = std::string(source_str, limit - length, length);
+    lines_found++;
+  }
+}
+
```
indutny: Two newlines

src/llv8.cc
```diff
@@ -352,19 +402,69 @@ std::string SharedFunctionInfo::ToString(Error& err) {
...
+void Script::GetLineColumnFromPos(int64_t pos, int64_t& line, int64_t& column, Error& err) {
```
indutny: 80 column limit.

src/llv8.h
```diff
@@ -131,7 +133,10 @@ class Script : public HeapObject {
   inline HeapObject Source(Error& err);
   inline HeapObject LineEnds(Error& err);
 
-  std::string GetLineColumnFromPos(int64_t pos, Error& err);
+  void GetLines(uint64_t start_line, std::string lines[],
+                          uint64_t line_limit, uint32_t &lines_found,
+                          Error& err);
+  void GetLineColumnFromPos(int64_t pos, int64_t& line, int64_t& column, Error& err);
```
indutny: 80 column?

src/llv8.h
```diff
@@ -365,11 +370,16 @@ class JSFrame : public Value {
...
   inline Value GetReceiver(int count, Error &err);
   inline Value GetParam(int slot, int count, Error &err);
-
+  
+  uint32_t GetSourceForDisplay(bool set_line, uint32_t line_start,
+                           uint32_t line_limit, std::string lines[],
+                           uint32_t& lines_found, Error& err);
   std::string Inspect(bool with_args, Error& err);
   std::string InspectArgs(JSFunction fn, Error& err);
+
```
indutny: Unnecessary newline.

src/llv8.cc
bmeck: fixed, check full diff view

src/llv8.cc
```diff
@@ -373,21 +475,18 @@ std::string Script::GetLineColumnFromPos(int64_t pos, Error& err) {
...
       line++;
     }
   }
-
-  snprintf(tmp, sizeof(tmp), ":%d:%d", static_cast<int>(line),
-      static_cast<int>(column));
-  return tmp;
 }
 
-
```
indutny: Erm? :)

src/llv8.cc
```diff
@@ -224,7 +271,6 @@ std::string JSFrame::InspectArgs(JSFunction fn, Error& err) {
   return res;
 }
 
-
```
indutny: Ditto.

indutny:
Landed in 92135c8. I think we need a bit better documentation for it!

Going to run `clang-format` on a project.

indutny:
Thank you!

PR #8, opened 2015-11-19
improving colors of backtrace command

thlorenz:
- using ansi escape codes to make JS trace blend better
- JS frame description in green to distinguish it from C/C++

screen shot 2015-11-19 at 1 16 48 pm

If you'd like a different color for the description, please LMK.
I figured green is good for JS (github uses it too).

### DIFF
--- a/src/llnode.cc
+++ b/src/llnode.cc
@@ -65,8 +65,8 @@ bool BacktraceCmd::DoExecute(SBDebugger d, char** cmd,
     if (err.Fail()) continue;
 
     // V8 symbol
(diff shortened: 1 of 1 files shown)


thlorenz:
We could also use cyan to make it blend perfectly, but then it's hard to tell what's JS and whats C++

thlorenz:
OK, not a good implementation.
We need to respect lldb settings instead of hardcoding.
Closing for now.

indutny:
It doesn't look like we could reuse lldb settings, but I think we could use `.llnode` file in user's folder. This should be generally more flexible.

Issue #11, opened 2016-06-09
Enable llnode to build against a local build of lldb.


hhellyer:
On this branch https://github.com/hhellyer/llnode/tree/build_with_latest_lldb I have updated llnode.gyp so that you can build llnode against a locally built version of lldb. (I did this so I can test out my LLDB API changes to add memory region information but it's generally useful and might help other people who hit weird API problems.)

Mac OSX has stricter linking than Linux and requires that you link the dynamic library against any other libraries it will use. Linux only needs the headers.

This difference means the build procedure on Linux only requires -Dlldb_dir to be updated to the location containing your lldb source. (If you have checked out the source it should be <llvm_dir>/tools/lldb.)

The LLVM build process places the built binaries out side the source tree so to build on Mac OSX you need to specify -Dlldb_dir (as on Linux) and a new parameter -Dlldb_build_dir. If you have followed the standard lldb build setup this is likely to be _llvm_dir_/../build.

-Dlldb_lib can still be used to override the name of the lldb library file but as it liblldb is generally soft-linked to liblldb.<version> it is usually safe to omit it.

(A side effect of these changes is that you don’t need to pass -Dlldb_lib on Linux any more.)

The Mac configure step, when using a local build of lldb is:

./gyp_llnode -Dlldb_dir=_some_path_/llvm/tools/lldb -Dlldb_build_dir=_some_path_/build

The Linux config step is:
./gyp_llnode -Dlldb_dir=_some_path_/llvm/tools/lldb

The normal build instructions are unchanged.

If this seems reasonable I'll create a pull request but it seemed better to discuss it here first.

indutny:
This seems very reasonable and nice! Thank you for an awesome work on this. Please file a PR 😉

indutny:
#12 has added this feature, thank you!

PR #12, opened 2016-06-13
Enable Mac to build llnode against a locally compiled version of lldb.

hhellyer:
This is for #11 - You've already done the change for Linux by removing the lib flag so I've updated the comment just to reference Mac.


indutny:
LGTM. Thank you!

indutny:
Landed in 368d27c407cf2fdca20a9f519d23439194b88d69, thank you again!

PR #13, opened 2016-06-17
Add support for listing all the JavaScript objects contained in a core dump

hhellyer:
This branch provides a command to list the objects in a dump grouped by type (map) and to dump all the instances of objects which share a particular map.

It needs the user to run one of the supplied scripts to generate a file of memory regions and them to export the name of that file in the RANGESFILE environment variable. (The readelf script should be run on linux, the otool one on Mac OSX.)

This restriction can be lifted once the lldb API is updated but only requires the code for walking memory to change as I’ve isolated that from the code which actually identifies objects. I think the walking memory code could probably be updated to improve performances as well as currently the memory is walked a word at a time.

Please feel free to suggest improvements and alterations or any bugs.

### DIFF
--- a/scripts/otool2segments.py
+++ b/scripts/otool2segments.py
@@ -1,0 +1,43 @@
+'''
+Created on 7 Apr 2016
+
+@author: hhellyer
+'''
+import sys,os,subprocess
+
+OTOOL_COMMAND = "otool -l {0}"
+
+# Grab the details of the memory ranges in the process from otool.
+def main():
+    if( len(sys.argv) < 2 ):
+        print("Usage " + sys.argv[0] + " <core_file>")
+        sys.exit(1)
+    core_file = sys.argv[1]
+
+    readelf_proc = subprocess.Popen(OTOOL_COMMAND.format(core_file), shell=True, bufsize=1,
+          stdin=subprocess.PIPE, stdout=subprocess.PIPE, close_fds=True)
+    (readelf_stdin, readelf_stdout) = (readelf_proc.stdin, readelf_proc.stdout)
+    
+    reading_segment = False
+    vaddress = ""
+
+    for line in readelf_stdout:
+        line = line.strip()
+        if not line.startswith("vmaddr") and not reading_segment:
+            continue
+        elif line.startswith("vmaddr"):
+            # Might need to filer out segments that have a file offset of 0 (ie aren't in the core!)
+            reading_segment = True
+            (name, vaddress) = line.split()
+        elif line.startswith("vmsize") and reading_segment:
+            reading_segment = False
+            memsize = line.split()[1]
+            # Simple format "address size", both in hex
+            print("{0} {1}".format(vaddress, memsize))
+            vaddress = ""
+    
+	
+
+
+if __name__ == '__main__':
+    main()
--- a/scripts/otool2segments.py
+++ b/scripts/otool2segments.py
@@ -14,19 +14,21 @@ def main():
         sys.exit(1)
     core_file = sys.argv[1]
 
-    readelf_proc = subprocess.Popen(OTOOL_COMMAND.format(core_file), shell=True, bufsize=1,
-          stdin=subprocess.PIPE, stdout=subprocess.PIPE, close_fds=True)
-    (readelf_stdin, readelf_stdout) = (readelf_proc.stdin, readelf_proc.stdout)
-    
+    otool_proc = subprocess.Popen(
+        OTOOL_COMMAND.format(core_file), shell=True, bufsize=1,
+        stdin=subprocess.PIPE, stdout=subprocess.PIPE, close_fds=True)
+    (otool_stdin, otool_stdout) = (otool_proc.stdin, otool_proc.stdout)
+
     reading_segment = False
     vaddress = ""
 
-    for line in readelf_stdout:
+    for line in otool_stdout:
         line = line.strip()
         if not line.startswith("vmaddr") and not reading_segment:
             continue
         elif line.startswith("vmaddr"):
-            # Might need to filer out segments that have a file offset of 0 (ie aren't in the core!)
+            # Might need to filer out segments that have a file offset of 0
+            # (ie segments that aren't in the core!)
             reading_segment = True
             (name, vaddress) = line.split()
         elif line.startswith("vmsize") and reading_segment:
@@ -35,9 +37,6 @@ def main():
             # Simple format "address size", both in hex
             print("{0} {1}".format(vaddress, memsize))
             vaddress = ""
-    
-	
-
 
 if __name__ == '__main__':
     main()
--- a/src/llnode.cc
+++ b/src/llnode.cc
@@ -7,6 +7,7 @@
 
 #include "src/llnode.h"
 #include "src/llv8.h"
+#include "src/llscan.h"
 
 namespace llnode {
 
@@ -281,6 +282,21 @@ bool PluginInitialize(SBDebugger d) {
                     "Syntax: v8 source list\n");
   interpreter.AddCommand("jssource", new llnode::ListCmd(),
                          "Alias for `v8 source list`");
+
+  v8.AddCommand("listobjects", new llnode::ListObjectsCmd(),
+      "List all object types and instance counts grouped by map and sorted by instance count.\n"
+      "Requires RANGESFILE to be set to a file containing memory ranges for the core file being debugged.");
+
+  interpreter.AddCommand("listobjects", new llnode::ListObjectsCmd(),
+       "Alias for `v8 listobjects`");
+
+  v8.AddCommand("listinstances", new llnode::ListInstancesCmd(),
+       "List all objects which share the specified map.\n"
+       "Accepts the same options as `v8 inspect`");
+
+  interpreter.AddCommand("listinstances", new llnode::ListInstancesCmd(),
+          "List all objects which share the specified map.\n");
+
 
   return true;
 }
--- a/src/llnode.cc
+++ b/src/llnode.cc
@@ -283,20 +283,20 @@ bool PluginInitialize(SBDebugger d) {
   interpreter.AddCommand("jssource", new llnode::ListCmd(),
                          "Alias for `v8 source list`");
 
-  v8.AddCommand("listobjects", new llnode::ListObjectsCmd(),
+  v8.AddCommand("findjsobjects", new llnode::FindObjectsCmd(),
                 "List all object types and instance counts grouped by map and "
                 "sorted by instance count.\n"
                 "Requires RANGESFILE to be set to a file containing memory "
                 "ranges for the core file being debugged.");
 
-  interpreter.AddCommand("listobjects", new llnode::ListObjectsCmd(),
-                         "Alias for `v8 listobjects`");
+  interpreter.AddCommand("findjsobjects", new llnode::FindObjectsCmd(),
+                         "Alias for `v8 findjsobjects`");
 
-  v8.AddCommand("listinstances", new llnode::ListInstancesCmd(),
+  v8.AddCommand("findjsinstances", new llnode::FindInstancesCmd(),
                 "List all objects which share the specified map.\n"
                 "Accepts the same options as `v8 inspect`");
 
-  interpreter.AddCommand("listinstances", new llnode::ListInstancesCmd(),
+  interpreter.AddCommand("findjsinstances", new llnode::FindInstancesCmd(),
                          "List all objects which share the specified map.\n");
 
 
--- a/src/llnode.cc
+++ b/src/llnode.cc
@@ -1,13 +1,13 @@
 #include <errno.h>
+#include <getopt.h>
 #include <stdlib.h>
 #include <string.h>
-#include <getopt.h>
 
 #include <lldb/API/SBExpressionOptions.h>
 
 #include "src/llnode.h"
-#include "src/llv8.h"
 #include "src/llscan.h"
+#include "src/llv8.h"
 
 namespace llnode {
 
@@ -284,18 +284,20 @@ bool PluginInitialize(SBDebugger d) {
                          "Alias for `v8 source list`");
 
   v8.AddCommand("listobjects", new llnode::ListObjectsCmd(),
-      "List all object types and instance counts grouped by map and sorted by instance count.\n"
-      "Requires RANGESFILE to be set to a file containing memory ranges for the core file being debugged.");
+                "List all object types and instance counts grouped by map and "
+                "sorted by instance count.\n"
+                "Requires RANGESFILE to be set to a file containing memory "
+                "ranges for the core file being debugged.");
 
   interpreter.AddCommand("listobjects", new llnode::ListObjectsCmd(),
-       "Alias for `v8 listobjects`");
+                         "Alias for `v8 listobjects`");
 
   v8.AddCommand("listinstances", new llnode::ListInstancesCmd(),
-       "List all objects which share the specified map.\n"
-       "Accepts the same options as `v8 inspect`");
+                "List all objects which share the specified map.\n"
+                "Accepts the same options as `v8 inspect`");
 
   interpreter.AddCommand("listinstances", new llnode::ListInstancesCmd(),
-          "List all objects which share the specified map.\n");
+                         "List all objects which share the specified map.\n");
 
 
   return true;
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -24,7 +24,7 @@ LLScan llscan;
 using namespace lldb;
 
 
-bool ListObjectsCmd::DoExecute(SBDebugger d, char** cmd,
+bool FindObjectsCmd::DoExecute(SBDebugger d, char** cmd,
                                SBCommandReturnObject& result) {
   SBTarget target = d.GetSelectedTarget();
   if (!target.IsValid()) {
@@ -41,28 +41,25 @@ bool ListObjectsCmd::DoExecute(SBDebugger d, char** cmd,
    * TODO(hhellyer) - Make sort type an option (by count, size or name)
    */
   std::vector<TypeRecord*> sorted_by_count;
-  std::map<uint64_t, TypeRecord*>::iterator end =
-      llscan.GetMapsToInstances().end();
-  for (std::map<uint64_t, TypeRecord*>::iterator it =
-           llscan.GetMapsToInstances().begin();
+  TypeRecordMap::iterator end = llscan.GetMapsToInstances().end();
+  for (TypeRecordMap::iterator it = llscan.GetMapsToInstances().begin();
        it != end; ++it) {
     sorted_by_count.push_back(it->second);
   }
 
   std::sort(sorted_by_count.begin(), sorted_by_count.end(),
-            TypeRecord::compareInstanceCounts);
+            TypeRecord::CompareInstanceCounts);
 
   uint64_t total_objects = 0;
 
-  result.Printf("Map                Instances  Total Size Name\n");
-  result.Printf("------------------ ---------- ---------- ----\n");
+  result.Printf(" Instances  Total Size Name\n");
+  result.Printf(" ---------- ---------- ----\n");
 
   for (std::vector<TypeRecord*>::iterator it = sorted_by_count.begin();
        it != sorted_by_count.end(); ++it) {
     TypeRecord* t = *it;
-    result.Printf("0x%016" PRIx64 " %10" PRId64 " %10" PRId64 " %s\n", t->map,
-                  t->instance_count, t->total_instance_size,
-                  t->type_name->c_str());
+    result.Printf(" %10" PRId64 " %10" PRId64 " %s\n", t->instance_count,
+                  t->total_instance_size, t->type_name->c_str());
     total_objects += t->instance_count;
   }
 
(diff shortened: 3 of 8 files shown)


src/llscan.cc
```diff
@@ -0,0 +1,470 @@
...
+
+struct MemoryRange {
+    uint64_t start;
+    uint64_t length;
+    MemoryRange* next;
+};
+
+static MemoryRange* ranges = nullptr;
+
+std::map<uint64_t, TypeRecord*>* mapstoinstances = NULL;
```
indutny: I suggest that we should use a `LLScan` class where `mapstoinstances` is a property. Command handling code could instantiate the class and pass it over to this file. Does it sound reasonable?

src/llscan.cc
```diff
@@ -0,0 +1,470 @@
...
+        return false;
+    }
+
+    /* Ensure we have a map of objects. */
+    if (!ScanHeapForObjects(&target, result)) {
+        return false;
+    }
+
+    /* Create a vector to hold the entries sorted by instance count
+     * TODO - Make sort type an option (by count, size or name)
```
indutny: Nitpicking, but let's put a name here (either me or you): `TODO(name)`.

src/llscan.cc
```diff
@@ -0,0 +1,470 @@
...
+    return "(ArrayBufferView)";
+  }
+
+  if (type == v8->types()->kJSDateType) {
+    return "(Date)";
+  }
+
+  std::string unknown("unknown: ");
+
+  return unknown += std::to_string(type);
```
indutny: Any reason in `+=` instead of `+` here?

src/llscan.cc
```diff
@@ -0,0 +1,470 @@
...
+  v8::LLV8 *v8 = heap_object->v8();
+  if (type == v8->types()->kGlobalObjectType) return heap_object->Inspect(options, err);
+  if (type == v8->types()->kCodeType) return heap_object->Inspect(options, err);
+  if (type == v8->types()->kMapType) {
+    return heap_object->Inspect(options, err);
+  }
+
+  if (type == v8->types()->kJSObjectType) {
+
+      llnode::v8::HeapObject map_obj = heap_object->GetMap(err);
```
indutny: Indent seems to be off here, but I'll fix it with `make format`.

src/llscan.cc
```diff
@@ -0,0 +1,470 @@
...
+
+  if (type == v8->types()->kJSDateType) {
+    return "(Date)";
+  }
+
+  std::string unknown("unknown: ");
+
+  return unknown += std::to_string(type);
+}
+
```
indutny: Two newlines, please.

src/llscan.cc
```diff
@@ -0,0 +1,470 @@
...
+                t->type_name->c_str());
+        total_objects += t->instance_count;
+    }
+
+    return true;
+}
+
+/* Utility function to generate short type names for objects.
+ * Based on the inspection code in llv8.cc
+ * TODO Merge back in to llv8.cc?
```
indutny: Same thing about `TODO(name)`

src/llscan.cc
```diff
@@ -0,0 +1,470 @@
...
+  }
+
+  if (type == v8->types()->kJSObjectType) {
+
+      llnode::v8::HeapObject map_obj = heap_object->GetMap(err);
+      if (err.Fail()) {
+          return std::string();
+      }
+
+      llnode::v8::Map map(map_obj);
```
indutny: I don't think that `llnode::` is need here or anywhere else, all functions are in `llnode` namespace already.

src/llscan.cc
```diff
@@ -0,0 +1,470 @@
...
+        result.Printf("No objects found with map %" PRIx64 "\n", mapaddr);
+    }
+
+    return true;
+
+}
+
+FindJSObjectsVisitor::FindJSObjectsVisitor(SBTarget* target,
+        SBCommandReturnObject* result) {
+    _target = target;
```
indutny: Nitpicking again, sorry! Convention is slightly different in this project. We use `target_` with `_` at the end instead. May I ask you to fix this?

src/llscan.cc
```diff
@@ -0,0 +1,470 @@
...
+
+    } else {
+        result.Printf("No objects found with map %" PRIx64 "\n", mapaddr);
+    }
+
+    return true;
+
+}
+
+FindJSObjectsVisitor::FindJSObjectsVisitor(SBTarget* target,
```
indutny: Should it be `SBTarget& target` instead?

indutny: Or even just `SBTarget`.

src/llscan.cc
```diff
@@ -0,0 +1,470 @@
...
+        SBCommandReturnObject* result) {
+    _target = target;
+    _result = result;
+    // Load V8 constants from postmortem data
+    llv8.Load(*target);
+    _address_byte_size = _target->GetProcess().GetAddressByteSize();
+    _found_count = 0;
+    mapstoinstances = new std::map<uint64_t, TypeRecord*>();
+}
+
```
indutny: Two newlines here, and everywhere between functions.

src/llscan.cc
```diff
@@ -0,0 +1,470 @@
...
+    }
+    if (heap_object.IsHoleOrUndefined(err)) {
+        return _address_byte_size;
+    }
+    if (err.Fail()) {
+        return _address_byte_size;
+    }
+
+    llnode::v8::HeapObject map_object = heap_object.GetMap(err);
+    if (err.Fail() || !map_object.Check()) {
```
indutny: I think `GetMap` calls `Check` automatically through the `LoadFieldValue<HeapObject>`.

src/llscan.cc
```diff
@@ -0,0 +1,470 @@
...
+// Defined in llnode.cc
+extern v8::LLV8 llv8;
+
+struct MemoryRange {
+    uint64_t start;
+    uint64_t length;
+    MemoryRange* next;
+};
+
+static MemoryRange* ranges = nullptr;
```
indutny: This should be an instance property too.

indutny: This whole thing may be instantiated globally in `llnode.cc`, just as `LLV8`. The same caching by comparing that `target_` to current should probably apply.

src/llscan.h
```diff
@@ -0,0 +1,89 @@
...
+  bool IsAHistogramType( llnode::v8::HeapObject *heap_object, v8::Error err );
+
+  lldb::SBTarget* _target;
+  lldb::SBCommandReturnObject* _result;
+  uint32_t _address_byte_size;
+  uint32_t _found_count;
+
+};
+
+std::string GetTypeName( llnode::v8::HeapObject *heap_object, v8::Value::InspectOptions *options, v8::Error err );
```
indutny: These four doesn't appear to be used anywhere outside of `llscan.cc`. Perhaps we could make them static and declare in `llscan.cc`?

indutny:
Left first serious of comments. Overall this thing looks fantastic! Thank you so much for your work on this!

May I ask you run `make format` on the code? It requires `clang-format` and should fix indents and possibly some other stuff. Thank you again.

hhellyer:
I've no problem making those changes, I was more worried about using the map id as the way of categorising objects by type and the code for generating short names for types. If you're happy with the overall approach then that's good news. I'm on vacation tomorrow but I'll get on with doing the tidy up when I return.

indutny:
@hhellyer this is indeed an important question! I suggest that a `constructor` may be used instead. There is not much need in differentiating objects by maps, and v8 may create tons of maps for a single constructor.

indutny:
@hhellyer considering that you are going on vacation, would you mind if I'll land your commit as it is and make some changes/fixes in a separate commits?

rnchamberlain:
@indutny - @hhellyer is only out for a day, but feel free to land the commit asis if you like

indutny:
@rnchamberlain oh, this changes everything! :) I'll wait for @hhellyer then, thank you!

hhellyer:
I've pushed a new update which (should) resolve most of the comments about code structure and formatting. I'll make the changes for how types are identified (names vs maps) in the next commit.
I'm afraid I labelled some of the TODOs with your name - mostly where I wanted a second opinion more than anything else.

src/llnode.cc
```diff
@@ -282,6 +283,23 @@ bool PluginInitialize(SBDebugger d) {
   interpreter.AddCommand("jssource", new llnode::ListCmd(),
                          "Alias for `v8 source list`");
 
+  v8.AddCommand("listobjects", new llnode::ListObjectsCmd(),
+                "List all object types and instance counts grouped by map and "
+                "sorted by instance count.\n"
+                "Requires RANGESFILE to be set to a file containing memory "
+                "ranges for the core file being debugged.");
+
+  interpreter.AddCommand("listobjects", new llnode::ListObjectsCmd(),
```
indutny: We probably want to call it `findjsobjects` or something like this. These aliases are for people who are used to `mdb_v8`.

src/llscan.cc
```diff
@@ -0,0 +1,497 @@
...
+    total_objects += t->instance_count;
+  }
+
+  return true;
+}
+
+
+/* Utility function to generate short type names for objects.
+ * Based on the inspection code in llv8.cc
+ * TODO(indutny) - Should this merge back in to llv8.cc?
```
indutny: I would live it here for now until the need for it will arise anywhere else.

src/llscan.cc
```diff
@@ -0,0 +1,497 @@
...
+  inspect_options.detailed = false;
+  inspect_options.print_map = false;
+  inspect_options.string_length = 0;
+
+  /* No entry in the map, create a new one. */
+  if (mapstoinstances_.count(map_object.raw()) == 0) {
+    TypeRecord* t = new TypeRecord();
+    t->map = map_object.raw();
+    t->instance_count++;
+    // TODO(indutny) - Is this correct or is there a better way to find the size
```
indutny: I think this is correct.

src/llscan.cc
```diff
@@ -0,0 +1,497 @@
...
+  }
+
+  if (err.Fail()) {
+    return address_byte_size_;
+  }
+
+  found_count_++;
+
+  /* Just advance one word.
+   * (Should advance by object size, assuming objects can't overlap!)
```
indutny: This will probably not work for now. The core file segment size is not necessarily a start of the heap, and if we will skip more than a pointer, we may eventually skip a lot of objects.

src/llscan.cc
```diff
@@ -0,0 +1,497 @@
...
+    /* Determine if this is a new instance.
+     * (We are scanning pointers to objects, we may have seen this location
+     * before.)
+     */
+    if (t->instances->count(word) == 0) {
+      t->instances->insert(word);
+      t->instance_count++;
+      // TODO(indutny) - Is this correct or is there a better way to find the
+      // size of one instance of an object?
+      t->total_instance_size += map.InstanceSize(err);
```
indutny: Ditto.

src/llscan.h
```diff
@@ -0,0 +1,113 @@
...
+  LLScan(){};
+
+  std::string GetTypeName(v8::HeapObject& heap_object,
+                          v8::Value::InspectOptions& options, v8::Error err);
+  bool ScanHeapForObjects(lldb::SBTarget target,
+                          lldb::SBCommandReturnObject result);
+  bool GenerateMemoryRanges(lldb::SBTarget target,
+                            const char* segmentsfilename);
+
+  std::map<uint64_t, TypeRecord*>& GetMapsToInstances() {
```
indutny: Should it be a `typedef`?

src/llscan.h
```diff
@@ -0,0 +1,113 @@
...
+  std::string* type_name;
+  uint64_t property_count;
+  uint64_t instance_count;
+  uint64_t total_instance_size;
+  std::set<uint64_t>* instances;
+
+  /* Sort records by instance count, use the other fields as tie breakers
+   * to give consistent ordering.
+   */
+  static bool compareInstanceCounts(TypeRecord* a, TypeRecord* b) {
```
indutny: Should be safe to use camel case here.

indutny: Gosh, I meant `CompareInstanceCounts`

src/llscan.cc
```diff
@@ -0,0 +1,497 @@
...
+
+  if (err.Fail()) {
+    result.SetError("Failed to evaluate expression");
+    return false;
+  }
+
+  uint64_t mapaddr = value.GetValueAsUnsigned();
+  std::map<uint64_t, TypeRecord*>::iterator instance_it =
+      llscan.GetMapsToInstances().find(mapaddr);
+  if (instance_it != llscan.GetMapsToInstances().end()) {
```
indutny: This is potentially going to print quite a lot of them. Should we introduce paging/limit options?

indutny: With some reasonable defaults, of course 😉

indutny:
@hhellyer That's ok, I have commented everywhere where you asked ;)

The changes are looking much better now, in fact I don't have any major comments for them at the moment. Please let me know when this will be ready for review.

Thank you!

src/llv8.h
```diff
@@ -403,6 +403,8 @@ class LLV8 {
 
   void Load(lldb::SBTarget target);
 
+  constants::Types types;
```
indutny: Should it use `friend class` instead?

src/llscan.cc
hhellyer: I don't know, lldb doesn't seem to do paging by default (unlike gdb) and doesn't seem to have an option to enable it so in general that's not the way an lldb command would work.
That said we may want to limit the output to make it readable later on but we could have options to only show the largest/most numerous types of objects rather than paging.
Can we pull this out and make it a future enhancement? :-)

indutny: Absolutely!

src/llv8.h
hhellyer: For that I have to add forward declarations for the class FindJSObjectsVisitor and the GetTypeName function (and it's parameters, one of which is another inner class) in llv8.h.
If GetTypeName moved though it would be simpler and we'd just need the forward declaration for FindJSObjectsVisitor.

indutny: Hm... I see it now. We should move `GetTypeName` to the `Value` class then. Thank you!

hhellyer:
I think I've made all the fixes now. The list is now indexed by type name and you should be able to do:
```
v8 findjsobjects
v8 findjsinstances <type>
v8 inspect <instance>
```
to get down to any particular object.

(As an aside my LLDB API changes have actually landed, I need to put some more changes in to actually back the API but getting the API fixed was the hard part as once in it can't change easily. Once that's in I'll update this code so it doesn't need the ranges file if the lldb support is there under another issue.)

scripts/otool2segments.py
```diff
@@ -0,0 +1,43 @@
...
+            # Might need to filer out segments that have a file offset of 0 (ie aren't in the core!)
+            reading_segment = True
+            (name, vaddress) = line.split()
+        elif line.startswith("vmsize") and reading_segment:
+            reading_segment = False
+            memsize = line.split()[1]
+            # Simple format "address size", both in hex
+            print("{0} {1}".format(vaddress, memsize))
+            vaddress = ""
+
```
indutny: There seems to be extra whitespace here, tabs, and 80 column limit violation in this and other python script. May I ask you to fix this, please?

src/llv8.h
```diff
@@ -447,6 +452,7 @@ class LLV8 {
   constants::Frame frame;
   constants::Types types;
 
+
```
indutny: Spurious change.

src/llv8.h
```diff
@@ -476,6 +482,7 @@ class LLV8 {
   friend class JSRegExp;
   friend class JSDate;
   friend class CodeMap;
+  friend class llnode::FindJSObjectsVisitor;
```
indutny: Is it still needed?

src/llv8.cc
```diff
@@ -672,6 +686,88 @@ std::string Smi::ToString(Error& err) {
...
+      return std::string();
+    }
+
+    int64_t constructor_type = constructor_obj.GetType(err);
+    if (err.Fail()) {
+      return std::string();
+    }
+
+    if (constructor_type != v8()->types()->kJSFunctionType) {
+      return "<Object: no constructor>";
```
indutny: Should it be just `(Object)` here for consistency?

src/llv8.cc
```diff
@@ -672,6 +686,88 @@ std::string Smi::ToString(Error& err) {
 }
 
 
+/* Utility function to generate short type names for objects.
+ */
+std::string HeapObject::GetTypeName(InspectOptions* options, Error& err) {
+  int64_t type = GetType(err);
+  if (type == v8()->types()->kGlobalObjectType) return Inspect(options, err);
```
indutny: I think it may be better to just return `(global)`, `(code)`, and `(map)` here.

src/llv8.cc
```diff
@@ -549,6 +549,20 @@ std::string Value::Inspect(InspectOptions* options, Error& err) {
 }
 
 
+std::string Value::GetTypeName(InspectOptions* options, Error& err) {
+  Smi smi(this);
+  if (smi.Check()) return smi.Inspect(err);
```
indutny: Mmm... is it used?

indutny: Looking at the function, I'd say that it should return something like "(Smi)" here.

src/llscan.h
```diff
@@ -0,0 +1,111 @@
...
+  LLScan(){};
+
+  std::string GetTypeName(v8::HeapObject& heap_object,
+                          v8::Value::InspectOptions& options, v8::Error err);
+  bool ScanHeapForObjects(lldb::SBTarget target,
+                          lldb::SBCommandReturnObject &result);
+  bool GenerateMemoryRanges(lldb::SBTarget target,
+                            const char* segmentsfilename);
+
+  TypeRecordMap& GetMapsToInstances() { return mapstoinstances_; };
```
indutny: `inline`?

src/llscan.h
```diff
@@ -0,0 +1,111 @@
...
+class MemoryVisitor {
+ public:
+  virtual ~MemoryVisitor(){};
+
+  virtual uint64_t Visit(uint64_t location, uint64_t available) { return 0; };
+};
+
+class TypeRecord {
+ public:
+  std::string* type_name;
```
indutny: I think these should be private properties with public `inline` getters.

Another thing is a `*` here. It doesn't look like these two fields need to be allocated separately. What do you think about  removing `*`, and also initializing all of them from the constructor function?

src/llscan.h
```diff
@@ -0,0 +1,111 @@
...
+ public:
+  ~FindObjectsCmd() override {}
+
+  bool DoExecute(lldb::SBDebugger d, char** cmd,
+                 lldb::SBCommandReturnObject& result) override;
+};
+
+class FindInstancesCmd : public CommandBase {
+ public:
+  ~FindInstancesCmd() override {}
```
indutny: Missing `;`?

src/llscan.h
```diff
@@ -0,0 +1,111 @@
...
+
+#include <lldb/API/LLDB.h>
+#include <map>
+#include <set>
+
+namespace llnode {
+
+class FindObjectsCmd : public CommandBase {
+ public:
+  ~FindObjectsCmd() override {}
```
indutny: Missing `;`?

src/llscan.h
```diff
@@ -0,0 +1,111 @@
...
+
+ private:
+  bool detailed_;
+};
+
+class MemoryVisitor {
+ public:
+  virtual ~MemoryVisitor(){};
+
+  virtual uint64_t Visit(uint64_t location, uint64_t available) { return 0; };
```
indutny: is there any reason to provide such default implementation, instead of doing `= 0` at the end?

src/llscan.cc
```diff
@@ -0,0 +1,376 @@
...
+    return false;
+  }
+
+  uint64_t address = 0, len = 0;
+
+  MemoryRange** tailptr = &ranges_;
+
+  lldb::addr_t address_byte_size = target.GetProcess().GetAddressByteSize();
+
+  while (input >> std::hex >> address >> std::hex >> len) {
```
indutny: Lovely! I'm still new to C++ streams :)

src/llscan.cc
```diff
@@ -0,0 +1,376 @@
...
+     * middle will be there, ranges are contiguous in the file, but cores often
+     * get truncated due
+     * to file size limits so ranges can be missing or truncated. Sometimes
+     * shared memory segments
+     * are omitted so it's also possible an entire section could be missing from
+     * the middle.
+     */
+    lldb::SBError error;
+    target.GetProcess().ReadPointerFromMemory(address, error);
+    if (error.Success()) {
```
indutny: Let's simplify this a bit:

```c
if (!error.Success()) {
  continue;
}
```

src/llscan.cc
```diff
@@ -0,0 +1,376 @@
...
+      target.GetProcess().ReadPointerFromMemory(
+          (address + len) - address_byte_size, error);
+
+      if (error.Success()) {
```
indutny: Same as above.

src/llscan.cc
```diff
@@ -0,0 +1,376 @@
...
+ */
+bool LLScan::GenerateMemoryRanges(lldb::SBTarget target,
+                                  const char* segmentsfilename) {
+  std::ifstream input(segmentsfilename);
+
+  if (!input.is_open()) {
+    return false;
+  }
+
+  uint64_t address = 0, len = 0;
```
indutny: I would prefer two separate declarations.

src/llscan.cc
```diff
@@ -0,0 +1,376 @@
...
+
+      if (error.Success()) {
+        MemoryRange* newRange = new MemoryRange();
+        if (nullptr == newRange) {
+          return false;
+        }
+        newRange->start = address;
+        newRange->length = len;
+        newRange->next = nullptr;
+        (*tailptr) = newRange;
```
indutny: No need in parens.

src/llscan.cc
```diff
@@ -0,0 +1,376 @@
...
+     */
+    lldb::SBError error;
+    target.GetProcess().ReadPointerFromMemory(address, error);
+    if (error.Success()) {
+      target.GetProcess().ReadPointerFromMemory(
+          (address + len) - address_byte_size, error);
+
+      if (error.Success()) {
+        MemoryRange* newRange = new MemoryRange();
+        if (nullptr == newRange) {
```
indutny: Can't fail AFAIK.

src/llscan.cc
```diff
@@ -0,0 +1,376 @@
...
+          return false;
+        }
+        newRange->start = address;
```
indutny: All of these could be initialized in constructor.

src/llscan.cc
```diff
@@ -0,0 +1,376 @@
...
+      head = head->next;
+      delete range;
+    }
+    ranges_ = nullptr;
+  }
+
+  /* Fall back to environment variable containing pre-parsed list of memory
+   * ranges. */
+  if (nullptr == ranges_) {
+    const char* segmentsfilename = getenv("RANGESFILE");
```
indutny: `RANGESFILE` sounds a bit generic. What do you think about prefixing it with `LLNODE_`?

src/llscan.cc
```diff
@@ -0,0 +1,376 @@
...
+   */
+
+  // Reload process anyway
+  process_ = target.GetProcess();
+
+  // Need to reload memory ranges (though this does assume the user has also
+  // updated
+  // RANGESFILE with data for the new dump or things won't match up).
+  if (target_ != target) {
+    MemoryRange* head = ranges_;
```
indutny: Let's put this to a separate method.

src/llscan.cc
```diff
@@ -0,0 +1,376 @@
...
+  }
+
+  /* Fall back to environment variable containing pre-parsed list of memory
+   * ranges. */
+  if (nullptr == ranges_) {
+    const char* segmentsfilename = getenv("RANGESFILE");
+
+    //
+    if (segmentsfilename == nullptr) {
+      //		  result.Printf("No memory ranges file. Plugin commands
```
indutny: No commented code, please.

src/llscan.cc
```diff
@@ -0,0 +1,376 @@
...
+    }
+    ranges_ = nullptr;
+  }
+
+  /* Fall back to environment variable containing pre-parsed list of memory
+   * ranges. */
+  if (nullptr == ranges_) {
+    const char* segmentsfilename = getenv("RANGESFILE");
+
+    //
```
indutny: Superfluous comment.

src/llscan.cc
```diff
@@ -0,0 +1,376 @@
...
+    /* Update an existing instance, if we haven't seen this instance before. */
+    TypeRecord* t = mapstoinstances_.at(type_name);
+    /* Determine if this is a new instance.
+     * (We are scanning pointers to objects, we may have seen this location
+     * before.)
+     */
+    if (t->instances->count(word) == 0) {
+      t->instances->insert(word);
+      t->instance_count++;
+      t->total_instance_size += map.InstanceSize(err);
```
indutny: Once these will become private members, we will need to have a method to do this accounting.

src/llscan.cc
```diff
@@ -0,0 +1,376 @@
...
+  inspect_options.detailed = false;
+  inspect_options.print_map = false;
+  inspect_options.string_length = 0;
+
+  std::string type_name = heap_object.GetTypeName(&inspect_options, err);
+
+  /* No entry in the map, create a new one. */
+  if (mapstoinstances_.count(type_name) == 0) {
+    TypeRecord* t = new TypeRecord();
+    t->type_name = new std::string(type_name);
```
indutny: This won't be required once constructor will be able to take these as params.

indutny:
Sorry, left lots of comments! The patch is looking really great.

src/llv8.h
hhellyer: I moved GetTypeName to be part of LLV8 which saves pre-declaring a messy function definition that's not part of any class in llv8.h but FindJSObjectsVisitor::IsAHistogramType still needs to look at the types to determine if we are interested in a given item on the heap.
I'm not sure I can sensibly move that function to llv8.h, it belongs with the visitor class, but I'm open to suggestions!

src/llv8.cc
hhellyer: Done.

src/llv8.cc
hhellyer: I've gone with (Global), (Code) and (Map) to match the capitalisation on the other names. Those in turn should match what users see on heap dumps but let me know if you'd rather I used lower case.

src/llscan.cc
hhellyer: I think you're right - `new` would probably throw an exception by default.

src/llscan.cc
hhellyer: Done, I haven't gone all the way and added accessors for all the fields of MemoryRange as the class is already a private inner class of LLScan - effectively part of the LLScan implementation.

src/llscan.cc
hhellyer: Done, I should have updated all the comments referring to it as well.

hhellyer:
I think I've gone through and done all the comments, I've pushed a new revision.

indutny:
Landed in c517e0ca09b714a487850ca74730d89f4a1c321d, thank you!

PR #16, opened 2016-07-12
Enable findjsobject/findjsinstances to use the SBMemoryRegionInfo API

hhellyer:
This patch enables LLNode to detect the SBMemoryRegionInfo API in LLDB at compile time. If it is present then LLNode will be compiled to use it and LLNODE_RANGESFILE won’t be needed in order to run findjsobjects.

Currently this will only work against a version of LLDB build from the current head of the LLDB development stream and the resulting plugin will only work with an lldb executable compiled from that.

I don’t know when the next level of LLDB, which hopefully should include the new API, will be released. I’m happy for this pull request to go on ice until then if you don't want support for unreleased function enabled yet - I really didn’t want it to only exist on my laptop!

### DIFF
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -277,6 +277,7 @@ bool LLScan::ScanHeapForObjects(lldb::SBTarget target,
 
 
 void LLScan::ScanMemoryRanges(FindJSObjectsVisitor& v) {
+  bool done = false;
 
 #if !defined(LLDB_SBMemoryRegionInfoList_h_)
   MemoryRange* head = ranges_;
@@ -303,6 +304,7 @@ void LLScan::ScanMemoryRanges(FindJSObjectsVisitor& v) {
       uint32_t increment =
           v.Visit(searchAddress, (address + len) - searchAddress);
       if (increment == 0) {
+        done = true;
         break;
       }
       searchAddress += increment;
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -238,7 +238,7 @@ bool LLScan::ScanHeapForObjects(lldb::SBTarget target,
     target_ = target;
   }
 
-#if !defined(LLDB_SBMemoryRegionInfoList_h_)
+#ifndef LLDB_SBMemoryRegionInfoList_h_
   /* Fall back to environment variable containing pre-parsed list of memory
    * ranges. */
   if (nullptr == ranges_) {
@@ -279,7 +279,7 @@ bool LLScan::ScanHeapForObjects(lldb::SBTarget target,
 void LLScan::ScanMemoryRanges(FindJSObjectsVisitor& v) {
   bool done = false;
 
-#if !defined(LLDB_SBMemoryRegionInfoList_h_)
+#ifndef LLDB_SBMemoryRegionInfoList_h_
   MemoryRange* head = ranges_;
 
   while (head != nullptr && !done) {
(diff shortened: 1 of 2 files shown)


src/llscan.cc
```diff
@@ -278,18 +277,28 @@ bool LLScan::ScanHeapForObjects(lldb::SBTarget target,
 
 
 void LLScan::ScanMemoryRanges(FindJSObjectsVisitor& v) {
-  MemoryRange* head = ranges_;
-
   bool done = false;
 
+#if !defined(LLDB_SBMemoryRegionInfoList_h_)
```
indutny: `ifndef`?

src/llscan.cc
```diff
@@ -278,18 +277,28 @@ bool LLScan::ScanHeapForObjects(lldb::SBTarget target,
...
+  MemoryRange* head = ranges_;
+
   while (head != nullptr && !done) {
     uint64_t address = head->start_;
     uint64_t len = head->length_;
     head = head->next_;
 
+# else
```
indutny: Could you please put `# else  // LLDB_SBMemoryRegionInfoList_h_` here.

src/llscan.cc
```diff
@@ -278,18 +277,28 @@ bool LLScan::ScanHeapForObjects(lldb::SBTarget target,
...
+  SBMemoryRegionInfoList memory_regions = process_.GetMemoryRegions();
+  SBMemoryRegionInfo region_info;
+
+  for( uint32_t i = 0; i < memory_regions.GetSize(); ++i ) {
+    memory_regions.GetMemoryRegionAtIndex(i, region_info);
+    uint64_t address = region_info.GetRegionBase();
+    uint64_t len = region_info.GetRegionEnd() - region_info.GetRegionBase();
+
+#endif
```
indutny: `#endif  // !LLDB_SBMemoryRegionInfoList_h_`

indutny:
I think we should land, but could you please give me some hints on how I could build this patched lldb locally (OS X)? A pointer to the relevant commit would be neat too!

hhellyer:
The commits into the llvm svn repository are here:

http://reviews.llvm.org/rL273547
http://reviews.llvm.org/rL274741

(There’s a Windows commit to come.)

The full build instructions are here:

http://lldb.llvm.org/build.html

but I ended up following the Linux cmake instructions on my Mac (I haven’t quite figured out if it was because I had never used Xcode before or the Xcode projects weren’t up to date but I couldn’t quite get it to work) so this is a summary of my notes, hopefully they’ll work for you! (If they don't work let me know, the lldb build isn't that hard once you've figured it out but it does take a while!)

Assuming you are working in $SOMEDIR

Create a build tree with lldb from the llvm git mirror:
> cd $SOMEDIR
> git clone http://llvm.org/git/llvm.git
> cd llvm/tools
> git clone http://llvm.org/git/clang.git
> git clone http://llvm.org/git/lldb.git

Configure and build:
> cd $SOMEDIR
> mkdir build
> cd build
> cmake $SOMEDIR/llvm -G "Unix Makefiles"
Build with:
> make -j8

[Wait a really long time while llvm, clang and lldb build.]

- You may need to follow the code signing instructions here: http://lldb.llvm.org/build.html#BuildingLldbOnMacOSX

Configure llnode with:
./gyp_llnode -Dlldb_dir=$SOMEDIR/llvm/tools/lldb -Dlldb_build_dir=$SOMEDIR/build

Run:
> $SOMEDIR/build/bin/lldb

and load your new plugin.

indutny:
Unfortunately, I don't have resources to test it at the moment. Landing as it is, I trust you! 👍

indutny:
Landed with lint and commit message fixes in 85a10aa. Thank you!

PR #15, opened 2016-07-09
minor fix for a coredump

brendangregg:

PR #18, opened 2016-07-19
PrintCmd - avoid crash when no options are passed to command.

hhellyer:
PR #15 fixed a bug in my findjsinstances parameter handling code which could cause a crash when no options were passed, unfortunately I used the same code as the v8 inspect command!
This change applies the same fix back to that and displays the help when no parameters are passed.


indutny:
Landed in 132701c with minor lint and commit message fixes. Thank you!

PR #19, opened 2016-08-08
nodeinfo command to show basic information about the process

hhellyer:
This PR uses the heap walk we are already doing to make it easy to view the contents of the process object. That lets us tell the user helpful things like what version of node they were running, what options and what script. (Output below.)

Doing this required adding some accessor functions based on the inspection code that let you navigate around objects rather than format them to strings. Those are in llv8.cc/.h.

The command has been added as “v8 nodeinfo” or just “nodeinfo” but I’m actually unsure whether it should be under “v8” at all as the process object is part of node.

I’m happy to make changes to the code or the output based on any feedback.

Sample output:
```
Information for process id 5780 (process=0x115a1f08469)
Platform = darwin, Architecture = x64, Node Version = v4.4.7
Component versions (process.versions=0x115a1f256b9):
    ares = 1.10.1-DEV
    http_parser = 2.5.2
    icu = 56.1
    modules = 46
    node = 4.4.7
    openssl = 1.0.2h
    uv = 1.8.0
    v8 = 4.5.103.36
    zlib = 1.2.8
Release Info (process.release=0x115a1f25729):
    name = node
    lts = Argon
    sourceUrl = https://nodejs.org/download/release/v4.4.7/node-v4.4.7.tar.gz
    headersUrl = https://nodejs.org/download/release/v4.4.7/node-v4.4.7-headers.tar.gz
Executable Path = [somedir]/node-v4.4.7-darwin-x64/bin/node
Command line arguments (process.argv=0x115a1f25831):
    [0] = '[somedir]/node-v4.4.7-darwin-x64/bin/node'
    [1] = '[somedir]/testscripts/print_info_and_crash.js'
Node.js Comamnd line arguments (process.execArgv=0x115a1f257f1):
    [0] = '--abort-on-uncaught-exception'
```

### DIFF
--- a/src/llnode.cc
+++ b/src/llnode.cc
@@ -311,6 +311,11 @@ bool PluginInitialize(SBDebugger d) {
   interpreter.AddCommand("findjsinstances", new llnode::FindInstancesCmd(),
                          "List all objects which share the specified map.\n");
 
+  v8.AddCommand("nodeinfo", new llnode::NodeInfoCmd(),
+                "Print information about Node.js\n");
+
+  interpreter.AddCommand("nodeinfo", new llnode::NodeInfoCmd(),
+                         "Print information about Node.js\n");
 
   return true;
 }
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -66,7 +66,6 @@ bool FindObjectsCmd::DoExecute(SBDebugger d, char** cmd,
   return true;
 }
 
-
 bool FindInstancesCmd::DoExecute(SBDebugger d, char** cmd,
                                  SBCommandReturnObject& result) {
   if (*cmd == NULL) {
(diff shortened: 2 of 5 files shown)


indutny:
@hhellyer sorry for such a big delay! I'm going to review it ASAP.

indutny:
@hhellyer may I ask you to format the commit message according to: https://github.com/indutny/llnode/blob/master/CONTRIBUTING.md#step-3-commit ?

We will clearly need to write a test for this, I'll help you with this as soon as we'll move past the formatting issue.

Thanks!

hhellyer:
@indutny - I think the commit message is straightened out. Let me know what format you want the testing to take and I'll add some test cases.

indutny:
Hooray! I'll do some prototyping on tests, should be trivial.

indutny:
Ah, nvm. It is not trivial at all.

Landing as it is for now, we should look into writing tests for both findjsobjects/findjsinstances and nodeinfo.

indutny:
Landed in b53010b, thank you!

hhellyer:
Thanks! I've done a blog post explaining it here:
https://developer.ibm.com/node/2016/08/25/nodeinfo-command-for-llnode/
(I wrote it with the command as that tends to make me more critical of what I'm doing.)
I'm happy to write some docs to check in to git too.

indutny:
@hhellyer looks cool!

Just noticed that we have added a global `nodeinfo` command. How intentional is this?

I guess we can keep it, but the future commands should be under `v8` namespace, unless we are mimicking the mdb's ones.

hhellyer:
It is under v8 as well. (Actually I was more worried about that since the process object is Node.js not v8 specific.)
I'm happy to remove the short cut for that and the other commands, let me know if you want a patch.

indutny:
Other shortcuts are to be "more compatible" with mdb_v8. This one has no analog in `mdb_v8`, so I think it would be great to remove the alias.

PR #21, opened 2016-08-25
llnode: remove nodeinfo shortcut

hhellyer:
The nodeinfo command has no equivalent in mdb_v8, removing the
shortcut and leaving it under the v8 namespace.


indutny:
LGTM

indutny:
Landed in 75d4a09e646b2f251ba3eae35119f16823c0e6d3, thank you!

Issue #22, opened 2016-08-31
CentOS 7 support


axot:
Compile will fail, here is the log output

```
../src/llv8.cc: In member function ‘std::string llnode::v8::JSFrame::Inspect(bool, llnode::v8::Error&)’:
../src/llv8.cc:245:43: error: expected ‘)’ before ‘PRIx64’
   snprintf(tmp, sizeof(tmp), " fn=0x%016" PRIx64, fn.raw());
```

I tried add #define __STDC_FORMAT_MACROS preprocess, then compile passed.
But, when I open lldb, `help v8` command get nothing subcommands.

```
(lldb) v8
The following subcommands are supported:

The following subcommands are supported:
```

My g++ version
```
$ g++ --version
g++ (GCC) 4.8.5 20150623 (Red Hat 4.8.5-4)
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

lldb version
```
$ lldb --version
lldb version 3.4.2 ( revision )
```

indutny:
Thank you for reporting this!

Does this patch fix the compilation error?

```diff
diff --git a/src/llscan.cc b/src/llscan.cc
index 2f9fb5a..5418d80 100644
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -1,9 +1,9 @@
 #include <errno.h>
-#include <inttypes.h>
 #include <stdlib.h>
 #include <string.h>
 
 #include <algorithm>
+#include <cinttypes>
 #include <fstream>
 #include <vector>
 
diff --git a/src/llv8.cc b/src/llv8.cc
index 5f878be..b6f60a2 100644
--- a/src/llv8.cc
+++ b/src/llv8.cc
@@ -1,5 +1,6 @@
 #include <assert.h>
-#include <inttypes.h>
+
+#include <cinttypes>
 
 #include "llv8-inl.h"
 #include "llv8.h"
```

Regarding `lldb`, what lldb source did you use when building `llnode`?

axot:
I has rebuild llnode with this patch, it works great! 
Im using lldb 3.4.2 when build llnode `./gyp_llnode -Dlldb_dir=/usr/lib64/llvm/`.
Is this version was too old?

```
$ ls /usr/lib64/llvm/
[...]

PR #23, opened 2016-08-31
src: use `cinttypes` to fix CentOS

indutny:
Otherwise it fails with:

    error: expected ‘)’ before ‘PRIx64’

Fix: #22


indutny:
cc @hhellyer @yjhjstz

yjhjstz:
LGTM !

indutny:
Landed in 8b7dca3, thank you!

PR #31, opened 2016-10-04
llv8: fix `WasNeutered` check

indutny:
The `kBitFieldOffset` was loaded conditionally, why it needs to be
computed for all versions of V8.

R= @hhellyer or @yjhjstz (PTAL)


indutny:
Landed in ad492ed, thank you!

PR #32, opened 2016-10-05
Add v8 findrefs command

hhellyer:
This PR adds a findrefs command to llnode to allow a user to search for all references to an object.

At the moment it’s just scanning the same set of objects that findjsobjects and findjsinstances use. That might miss references from v8 internals but does limit the results to things a JavaScript programmer has direct control over. That may need to change later on.

I’ve got a blog posting explaining the command ready to go, I’ll update it before posting with any feedback and changes to the command. I’ll also do an update to the v8 usage in README.md afterwards just to reflect the current contents of the help but the number of commands is starting to build up a bit. Would it be worth discussing (under a separate issue) about where in the project is the right place to put some more documentation and what it might need?

### DIFF
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -81,6 +81,12 @@ bool FindInstancesCmd::DoExecute(SBDebugger d, char** cmd,
     return false;
   }
 
+  /* Ensure we have a map of objects. */
+  if (!llscan.ScanHeapForObjects(target, result)) {
+    result.SetStatus(eReturnStatusFailed);
+    return false;
+  }
+
   v8::Value::InspectOptions inspect_options;
 
   inspect_options.detailed = detailed_;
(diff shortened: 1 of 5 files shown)


src/llscan.cc
```diff
@@ -289,6 +295,173 @@ bool NodeInfoCmd::DoExecute(SBDebugger d, char** cmd,
...
+        // result.Printf("Unhandled type: %" PRId64 " for addr %" PRIx64
+        //    "\n", type, addr);
+      }
+    }
+  }
+
+  result.SetStatus(eReturnStatusSuccessFinishResult);
+  return true;
+}
+
```
indutny: Double-newline, please. (Here and in other places between functions).

src/llscan.cc
```diff
@@ -289,6 +295,173 @@ bool NodeInfoCmd::DoExecute(SBDebugger d, char** cmd,
...
+  v8::Value search_value(&llv8, value.GetValueAsSigned());
+  v8::Smi smi(search_value);
+  if (smi.Check()) {
+    result.SetError("Search value is an SMI.");
+    return false;
+  }
+
+  // Walk all the object instances and handle them according to their type.
+  TypeRecordMap mapstoinstances = llscan.GetMapsToInstances();
+  for (auto const entry : mapstoinstances) {
```
indutny: Do we want to have a limit on this? Some objects may have way too many references.

src/llscan.cc
```diff
@@ -289,6 +295,173 @@ bool NodeInfoCmd::DoExecute(SBDebugger d, char** cmd,
...
+                                  v8::JSObject& js_obj, v8::Value& search_value,
+                                  v8::Error& err) {
+  v8::Value::InspectOptions inspect_options;
+  inspect_options.detailed = false;
+  int64_t length = js_obj.GetArrayLength(err);
+  for (int64_t i = 0; i < length; ++i) {
+    v8::Value v = js_obj.GetArrayElement(i, err);
+    if (err.Success() && v.raw() == search_value.raw()) {
+      std::string type_name = js_obj.GetTypeName(&inspect_options, err);
+      result.Printf("0x%" PRIx64 ": %s[%" PRId64 "]=0x%" PRIx64 "\n",
```
indutny: This could be more generic: instead of printing reference in-line - the callback may be passed to the `PrintRefs` functions. I'm not sure what arguments it should have, though... Perhaps `std::string` with object description and a `v8::Value`?

Doing it in a such way may be helpful in the future, if we would want to create tools for complicated analysis.

src/llv8.cc
```diff
@@ -1311,6 +1311,86 @@ void JSObject::Keys(std::vector<std::string>& keys, Error& err) {
...
+
+    if (err.Fail()) return;
+
+    // Skip holes
+    bool is_hole = key.IsHoleOrUndefined(err);
+    if (err.Fail()) return;
+    if (is_hole) continue;
+
+    Value value = dictionary.GetValue(i, err);
+
```
indutny: Superfluous newline.

src/llv8.cc
```diff
@@ -1311,6 +1311,86 @@ void JSObject::Keys(std::vector<std::string>& keys, Error& err) {
...
+  if (is_dict) {
+    DictionaryEntries(entries, err);
+  } else {
+    DescriptorEntries(entries, map, err);
+  }
+
+  return;
+}
+
+void JSObject::DictionaryEntries(std::vector<std::pair<Value, Value>>& entries,
```
indutny: Should these just return `std::vector<>`?

src/llv8.cc
```diff
@@ -1311,6 +1311,86 @@ void JSObject::Keys(std::vector<std::string>& keys, Error& err) {
...
+  bool is_dict = map.IsDictionary(err);
+  if (err.Fail()) return;
+
+  if (is_dict) {
+    DictionaryEntries(entries, err);
+  } else {
+    DescriptorEntries(entries, map, err);
+  }
+
+  return;
```
indutny: Superfluous return.

src/llv8.cc
hhellyer: You mean instead of being passed a vector by reference? (Not being passed something of type std::vector<>? I may have misunderstood this comment.)

I *think* it probably involves less copying as it is (C++ is good at catching me out though) and I added these functions to get round a performance issue so that was the focus when I put them in. It is a slight nuisance for the caller so possibly making the top level public function return a std::vector and the protected Dictionary and Descriptor functions use one they are given might hide the implementation better.

src/llscan.cc
hhellyer: I considered putting the all the PrintRefs functions on a callback class.
Their interface could remain as is (moved into a new class) but the search_value could be then be treated as the referred to object, the property name to search for, or a string value.
FindReferencesCmd would then instantiate the right one based on the flags it was given.

hhellyer:
I've pushed changes which should resolve all the basic issues.
I've restructured how the PrintRefs code works to make it easier to support other types of search.
I think there are still open questions on whether I add an option for handling too many results now or later and what the return value from the JSObject::Entries functions should be.

src/llscan.cc
hhellyer: I think you are probably right - Can I raise a separate issue though? It would probably happen when there are a lot of objects of the same type and the same scenarios would hit findjsinstances in exactly the same way. I can add one consistent option to both the commands at the same time.

src/llv8.cc
indutny: It should be fine and a bit easier to understand. Are you sure there were any performance issues because of copying?

src/llscan.cc
indutny: Yeah, something like this. I think we can do it later once this is landed.

src/llv8.cc
hhellyer: No, the performance issues were due to creating JSStrings when they weren't actually needed. That was fairly catastrophic. (I was creating a std::string out of every property name rather than just the ones that pointed to the value we were interested in.)

I made the assumption that it would be better to pass by ref but it looks like that was wrong. I did some basic wall clock timing of the two approaches on the core I used to re-create the original problem they actually both came out the same so I'll push the change.

Either copying the vector isn't that big a deal, the compiler is optimising it or the vector implementation is a bit clever about how it's copied. (Or any combination of those!)

src/llv8.cc
```diff
@@ -1344,8 +1422,6 @@ void JSObject::DictionaryKeys(std::vector<std::string>& keys, Error& err) {
   int64_t length = dictionary.Length(err);
   if (err.Fail()) return;
 
-  Value::InspectOptions options;
```
indutny: Oh, unused variable, nice!

PR #33, opened 2016-10-07
Add —-print-source (-s) flag

hhellyer:
This adds the option to display the full source associated with any function object via a --print-source (-s) flag when running v8 inspect. I think the choice of flag name makes sense but I'm happy to change it.

Some sample output is below. I'm happy to change the format of the output, though I can't do much about the formatting on the code that's being displayed. :-)

Sample output:
(lldb) v8 inspect --print-source 0x000022689a088939
0x000022689a088939:<function: helloworld at callfunctions.js:13:20
  context=0x000022689a0888a9
  source:
function helloworld() {
	console.log("Hello World!");
}
>

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Build Status](https://secure.travis-ci.org/indutny/llnode.png)](http://travis-ci.org/indutny/llnode)
 
-Node.js v4.x-v6.x C++ plugin for LLDB.
+Node.js v4.x-v6.x C++ plugin for [LLDB](http://lldb.llvm.org) - a next generation, high-performance debugger.
 
 ## Demo
 
--- a/src/llv8-inl.h
+++ b/src/llv8-inl.h
@@ -247,6 +247,15 @@ inline int64_t SharedFunctionInfo::StartPosition(Error& err) {
   return field;
 }
 
+// TODO (hhellyer): as above, this field is different on 32bit.
+inline int64_t SharedFunctionInfo::EndPosition(Error& err) {
+  int64_t field = LoadField(v8()->shared_info()->kEndPositionOffset, err);
+  if (err.Fail()) return -1;
+
+  field &= 0xffffffff;
+  return field >> 1;
+}
+
 ACCESSOR(JSFunction, Info, js_function()->kSharedInfoOffset,
          SharedFunctionInfo);
 ACCESSOR(JSFunction, GetContext, js_function()->kContextOffset, HeapObject);
--- a/src/llv8.cc
+++ b/src/llv8.cc
@@ -1394,6 +1394,84 @@ void JSObject::Keys(std::vector<std::string>& keys, Error& err) {
   return;
 }
 
+
+std::vector<std::pair<Value, Value>> JSObject::Entries(Error& err) {
+  HeapObject map_obj = GetMap(err);
+
+  Map map(map_obj);
+
+  bool is_dict = map.IsDictionary(err);
+  if (err.Fail()) return {};
+
+  if (is_dict) {
+    return DictionaryEntries(err);
+  } else {
+    return DescriptorEntries(map, err);
+  }
+}
+
+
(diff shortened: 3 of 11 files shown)


src/llv8.cc
```diff
@@ -313,12 +313,95 @@ std::string JSFunction::Inspect(InspectOptions* options, Error& err) {
...
+    return std::string();
+  }
+
+  // There is no `Script` for functions created in C++ (and possibly others)
+  int64_t type = script.GetType(err);
+  if (err.Fail()) {
+    return std::string();
+  }
+
+  if (type != v8()->types()->kScriptType) {
```
indutny: I wish there was a way to de-duplicate this code.

indutny:
Thank you!

hhellyer:
I was just looking at adding a test case. I'll try to add that to this branch tomorrow (unless you want to merge as is and I'll add it under another PR).

indutny:
Oh, right! Let's add test to this PR before landing it. Thanks!

hhellyer:
Test case added, since it runs extra tests on node v5 and above I've tried the checked it against a v4 node build and a v6 build and it seems to work for both.

src/llv8-inl.h
```diff
@@ -247,6 +247,15 @@ inline int64_t SharedFunctionInfo::StartPosition(Error& err) {
   return field;
 }
 
+// TODO (hhellyer): as above, this field is different on 32bit.
+inline int64_t SharedFunctionInfo::EndPosition(Error& err) {
+  int64_t field = LoadField(v8()->shared_info()->kEndPositionOffset, err);
+  if (err.Fail()) return -1;
+
+  field &= 0xffffffff;
+  return field >> 1;
```
indutny: Oh, it is a pseudo-smi. Cool!

test/inspect-test.js
```diff
@@ -90,24 +90,39 @@ tape('v8 inspect', (t) => {
...
-    let match = lines.match(
-        /\(closure\)=(0x[0-9a-f]+)[^\n]+function: closure/i);
-    t.ok(match, '`method` should have `closure`');
 
-    sess.send(`v8 inspect ${match[1]}`);
+    // Include 'source:' and '>' to act as boundaries. (Avoid
+    // passing if the whole file it displayed instead of just
+    // the function we want.)
+    const methodSource = "  source:\n" +
+    "  Class.prototype.method = function method() {\n" +
```
indutny: Should it print starting from `function method() {` instead? I think printing it this way may be problematic for minimified sources, where everything is on a single line.

hhellyer: It turns out the start position of the function source code in the script is the beginning of the arguments list at "(" not the function keyword or the (optional) function name. That makes sense but left the output looking a bit odd so I've inserted the function keyword and the function name (if there is one) ahead of the arguments list for readability.
The output expected by the test is now:

```
    const methodSource = "  source:\n" +
    "function method() {\n" +
    "    throw new Error('Uncaught');\n" +
    "  }\n" +
    ">"
```

indutny:
Great! Could you please add a test for an arrow function too?

hhellyer:
I've added the extra test case.
Once this is merged shall I do an update to README.md just to update the help v8 output? It's looking a little out of date.

README.md
```diff
@@ -2,7 +2,7 @@
 
 [![Build Status](https://secure.travis-ci.org/indutny/llnode.png)](http://travis-ci.org/indutny/llnode)
 
-Node.js v4.x-v6.x C++ plugin for LLDB.
+Node.js v4.x-v6.x C++ plugin for [LLDB](http://lldb.llvm.org) - a next generation, high-performance debugger.
```
indutny: Huh? I think this should not be here.

indutny:
Looks like your branch got all mixed up, otherwise LGTM.

Regarding readme, let's update it with `help v8`. This is what it is right now anyway.

indutny:
Yep, thanks!

PR #34, opened 2016-10-10
readme: add short description of LLDB

watson:
For those who don't know what LLDB is, this will make it a little easier to find out


indutny:
Thank you!

PR #35, opened 2016-10-10
readme: update Travis CI badge to reflect new github org

watson:
For some reason the current badge doesn't work anymore. I'm not a hundred percent sure this is related to the move to the nodejs org, but in any case I guess this should be updated.

I've also modified the URL slightly to reflect the latest conventions. Hopefully that's ok :)

As far as I can see the builds have not been enabled on Travis yet, so this change will not work before someone on the Node.js org does that I guess.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # llnode
 
-[![Build Status](https://secure.travis-ci.org/indutny/llnode.png)](http://travis-ci.org/indutny/llnode)
+[![Build Status](https://travis-ci.org/nodejs/llnode.svg)](http://travis-ci.org/nodejs/llnode)
 
 Node.js v4.x-v6.x C++ plugin for LLDB.
 


indutny:
I think the problem is that Travis is not enabled for llnode in this github org. cc @nodejs/ctc or @nodejs/collaborators , I have no idea who is in charge of our Travis account.

phillipj:
Afraid it's not as simple as just enabling Travis, due to org security
concerns. See the discussion in https://github.com/nodejs/help/issues/243

indutny:
Oh, right! I totally forgot about that. Should we hook it up to jenkins then?

phillipj:
@jbergstroem got any thoughts on this? Maybe we could setup a jenkins build specifically for llnode and possibly make the github-bot automatically kick off a build as a proof of concept for doing not using Travis?

rvagg:
Set up a quick Jenkins job to see if it can be done without too much pain but have stalled and need some input.

Here's my setup, I'm using one of the hosts that run Docker on it so we can containerise all of the builds and you can test whatever fork you like (you should do some basic code checks first though of course).

The container is setup as:

```
FROM ubuntu:trusty
RUN apt-get update
RUN apt-get install build-essential lldb-3.6 lldb-3.6-dev curl apt-transport-https git -y
RUN addgroup --gid 1000 iojs
RUN adduser iojs --uid 1000 --gid 1000 --gecos iojs --disabled-password
```

Then in Jenkins @ https://ci.nodejs.org/job/llnode/ (@indutny you _should_ have admin access to this if for now I've set it up right, so you can fiddle with 'configure') the script goes like this:

```
docker run -i --rm llnode:base /bin/bash -c '

curl -sL https://deb.nodesource.com/setup_7.x | bash -
apt-get install nodejs -y

cd /home/iojs && su iojs -c "

git clone https://github.com/'${github_user}'/'${github_repo}'.git llnode
cd llnode
git checkout '${github_ref}' -f
git clone https://chromium.googlesource.com/external/gyp.git tools/gyp
npm install
make _travis

"

'
```

The idea being that it can run multiple versions of Node just by rerunning this script with a different install version.

It runs, e.g. https://ci.nodejs.org/job/llnode/2/console, but fails with one of these for each test:

```
error: process launch failed: Child ptrace failed.
```

What's missing?

indutny:
@rvagg not sure how acceptable for CI it is, but I believe that `lldb` won't work if container doesn't run with `--privileged` flag. The reason for this is that `ptrace` syscall is not allowed by default, and it is the main primitive that is used for debugging on linuxes.

indutny:
@rvagg `--security-opt=seccomp:unconfined` works too.

indutny:
See https://github.com/docker/docker/issues/7276#issuecomment-194137403

indutny:
Or we may want to devise a custom security profile: https://docs.docker.com/engine/security/seccomp/

rvagg:
OK, well it's sort of working and you're aware of the getopts problem and I'll assume you're dealing with that. Job is @ https://ci.nodejs.org/job/llnode/4/console and you can trigger it with any fork & branch/ref on GitHub. It's currently viewable by all but executable only by you (@indutny). We don't have a @nodejs/llnode so I can't put a group in and there's only a couple of individuals with write access here. If you want the list of people expanded we should either switch to a group or do add them manually in the Jenkins job.

Since we are running a little less restricted than normal you just need to make sure you pass your eyes over PRs from untrusted sources to make sure there's nothing malicious being run on that server.

indutny:
@rvagg thank you!

indutny:
Something fishy is going on there: https://gist.github.com/indutny/e118dc1f11383ca77266fe32e4649c5b . I suspect this is another instance of lldb-3.6 headers vs binary incompatibility.

rvagg:
FYI to people reading this, we solved it by moving from Trusty to Xenial, something about older headers on Trusty (which might be worth checking out cause you'll likely get people wanting to run this on older distros).

rnchamberlain:
It would be good to test against lldb-3.9 if poss, it is available on xenial.

indutny:
May be a good idea. Right now I'm struggling with Ubuntu builds, it seems to be working pretty badly even with lldb-3.8 , will investigate more into it.

indutny:
The particular problem that I see when running `npm test` on Ubuntu vbox (and CI):

```
llnode: invalid option -- '9'
llnode: invalid option -- '�'
llnode: invalid option -- 't'
llnode: invalid option -- ''
```

rnchamberlain:
@indutny fyi I'm also seeing something similar, intermittently, using llnode on lldb 3.9, ubuntu 16.04 eg
```
(lldb) jsprint 0x00001eba885c3cb1
llnode: invalid option -- '▒'
llnode: invalid option -- 'x'
llnode: invalid option -- '▒'
llnode: invalid option -- 'R'
llnode: invalid option -- '<'
error: error: error: No value
```
I run it again and it works ok:
```
(lldb) jsprint 0x00001eba885c3cb1
0x00001eba885c3cb1:<Object: NativeModule properties {
    .filename=0x00001eba885c3cf1:<String: "_stream_readable...">,
    .id=0x00001507377e14a9:<String: "_stream_readable">,
    .exports=0x00001eba885cb9f9:<function: Readable at _stream_readable.js:104:18>,
    .loaded=0x00001507377043c1:<true>,
    .loading=0x0000150737704271:<false>}>
```

indutny:
@rnchamberlain I tend to believe that for some reason `std::string`s passed to llnode by lldb are freed before they are actually passed. Still have no idea why.

@bnoordhuis I know you have pretty solid experience in this area. If you have a minute - could you please help me figure out the cause of the issue here? I suspect it is binary/headers incompatibility, but not absolutely sure.

indutny:
Other possibility may be a `getopt_long` bug, or our misuse of it.

indutny:
Update: after reading `getopt` source, I think we are misusing it. I have a patch that may make it work, though.

In a long-term I suggest to move to a different library, or write a specialized lib ourselves.

indutny:
Here is PR: https://github.com/nodejs/llnode/pull/57

ChALkeR:
Btw, https://developer.github.com/early-access/integrations/ should fix the security concerns about granting access to Travis, so when/if Travis adopts that, it should be good.

cjihrig:
Can this be closed since https://github.com/nodejs/llnode/pull/107 landed?

bnoordhuis:
Yes, let's close.  Thanks anyway, @watson.

PR #39, opened 2016-10-17
readme: specify llnode clone

gibfahn:
Specify that you have to clone lldb before building

This should (I guess) be obvious, but it should probably be specified anyway.


rnchamberlain:
Change looks good, but the commit comment looks wrong, you have to clone llnode, not lldb.

gibfahn:
@rnchamberlain Whoops, fixed

PR #40, opened 2016-10-20
Add findrefs -n to search by property name

hhellyer:
This change adds a -n <name> option to v8 findrefs to search for all references referred to from properties named <name>. It also adds a -v option that matches the current (and still default with no flag specified) behaviour where all references to a property with that value are returned.

When findrefs was added ( https://github.com/nodejs/llnode/pull/32 ) I refactored it’s code as part of the review so that it could be easily extended. This enhancement takes advantage of that.

The output is the same as findrefs <object address> so the user sees the referring object address, it’s type, the field name and the object referred to from that property. Effectively it is scanning for all objects that contain a property with a particular name but the output is quite concise and allows the user to immediately inspect the object they are interested 

e.g:
(lldb) v8 findrefs --name constructor
0x2b57415b1211: (Array).constructor=0x2b57415b0f99
0x2f0367543b79: FastBuffer.constructor=0x2f036753dfa9
0x2f03675e2a51: KnownObject.constructor=0x2f03675e29f1
0x1d8468cdde11: Object.constructor=0x2b57415af6c1
0x1d8468cdde81: Object.constructor=0x2b57415cfc21
…

Searching for properties by name produces even more output than by value. I will look at adding the output limit we also discussed in https://github.com/nodejs/llnode/pull/32 next.

### DIFF
--- a/src/llnode.cc
+++ b/src/llnode.cc
@@ -323,7 +323,12 @@ bool PluginInitialize(SBDebugger d) {
                 "Print information about Node.js\n");
 
   v8.AddCommand("findrefs", new llnode::FindReferencesCmd(),
-                "Find all the objects that refer to the specified object.\n");
+                "Finds all the object properties which meet the search criteria.\n"
+                "The default is to list all the object properties that reference the specified value.\n"
+      "Flags:\n\n"
+      " * -v, --value expr     - all properties that refer to the specified JavaScript object (default)\n"
+      " * -n, --name  name     - all properties with the specified name\n"
+      "\n");
 
   return true;
 }
--- a/src/llnode.cc
+++ b/src/llnode.cc
@@ -322,11 +322,14 @@ bool PluginInitialize(SBDebugger d) {
   v8.AddCommand("nodeinfo", new llnode::NodeInfoCmd(),
                 "Print information about Node.js\n");
 
-  v8.AddCommand("findrefs", new llnode::FindReferencesCmd(),
-                "Finds all the object properties which meet the search criteria.\n"
-                "The default is to list all the object properties that reference the specified value.\n"
+  v8.AddCommand(
+      "findrefs", new llnode::FindReferencesCmd(),
+      "Finds all the object properties which meet the search criteria.\n"
+      "The default is to list all the object properties that reference the "
+      "specified value.\n"
       "Flags:\n\n"
-      " * -v, --value expr     - all properties that refer to the specified JavaScript object (default)\n"
+      " * -v, --value expr     - all properties that refer to the specified "
+      "JavaScript object (default)\n"
       " * -n, --name  name     - all properties with the specified name\n"
       "\n");
 
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -312,9 +312,9 @@ bool FindReferencesCmd::DoExecute(SBDebugger d, char** cmd,
   }
 
   // Default scan type.
-  ScanType type = ScanType::fieldvalue;
+  ScanType type = ScanType::kFieldValue;
 
-  char** start = ParseScanOptions(cmd, type);
+  char** start = ParseScanOptions(cmd, &type);
 
   if (*start == nullptr) {
     result.SetError("Missing search parameter");
@@ -328,7 +328,7 @@ bool FindReferencesCmd::DoExecute(SBDebugger d, char** cmd,
   ObjectScanner* scanner;
 
   switch (type) {
-    case ScanType::fieldvalue: {
+    case ScanType::kFieldValue: {
       std::string full_cmd;
       for (; start != nullptr && *start != nullptr; start++) full_cmd += *start;
 
@@ -353,18 +353,18 @@ bool FindReferencesCmd::DoExecute(SBDebugger d, char** cmd,
       scanner = new ReferenceScanner(search_value);
       break;
     }
-    case ScanType::propertyname: {
+    case ScanType::kPropertyName: {
       // TODO - Should I need to check that start++ is null?
       std::string property_name = *start;
       scanner = new PropertyScanner(property_name);
       break;
     }
-      /* We can add options to the command and further sub-classes of
-       * object scanner to do other searches, e.g.:
-       * - Objects that refer to a particular string literal.
-       *   (lldb) findreferences -s "Hello World!"
-       */
-    case ScanType::badoption: {
+    /* We can add options to the command and further sub-classes of
+     * object scanner to do other searches, e.g.:
+     * - Objects that refer to a particular string literal.
+     *   (lldb) findreferences -s "Hello World!"
+     */
+    case ScanType::kBadOption: {
       result.SetError("Invalid search type");
       result.SetStatus(eReturnStatusFailed);
       return false;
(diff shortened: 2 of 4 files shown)


rnchamberlain:
Just a note, further use of getopt makes the port to Windows harder, but we'll need a general solution for that anyway.

src/llnode.cc
```diff
@@ -323,7 +323,12 @@ bool PluginInitialize(SBDebugger d) {
                 "Print information about Node.js\n");
 
   v8.AddCommand("findrefs", new llnode::FindReferencesCmd(),
-                "Find all the objects that refer to the specified object.\n");
+                "Finds all the object properties which meet the search criteria.\n"
+                "The default is to list all the object properties that reference the specified value.\n"
```
indutny: 80 column limit? Does `make format` fix or complain about it?

src/llnode.cc
```diff
@@ -323,7 +323,12 @@ bool PluginInitialize(SBDebugger d) {
...
+      "Flags:\n\n"
```
indutny: Should it be aligned with the previous line?

src/llscan.h
```diff
@@ -41,12 +41,13 @@ class FindReferencesCmd : public CommandBase {
   bool DoExecute(lldb::SBDebugger d, char** cmd,
                  lldb::SBCommandReturnObject& result) override;
 
- private:
-  bool detailed_;
+  enum ScanType { fieldvalue, propertyname, badoption };
```
indutny: `kFieldValue, kPropertyName, kBadOption`

src/llscan.h
```diff
@@ -65,6 +66,20 @@ class FindReferencesCmd : public CommandBase {
    private:
     v8::Value& search_value_;
   };
+
+
+  class PropertyScanner : public ObjectScanner {
+   public:
+    PropertyScanner(std::string search_value) : search_value_(search_value){};
```
indutny: `) {}`

src/llv8.cc
```diff
@@ -1462,24 +1463,62 @@ std::vector<std::pair<Value, Value>> JSObject::DescriptorEntries(Map map, Error&
...
   std::vector<std::pair<Value, Value>> entries;
   for (int64_t i = 0; i < own_descriptors_count; i++) {
     Smi details = descriptors.GetDetails(i, err);
-    if (err.Fail()) return entries;
+    if (err.Fail()) continue;
 
     Value key = descriptors.GetKey(i, err);
-    if (err.Fail()) return entries;
+    if (err.Fail()) continue;
+
```
indutny: Superfluous newline.

src/llv8.cc
```diff
@@ -1462,24 +1463,62 @@ std::vector<std::pair<Value, Value>> JSObject::DescriptorEntries(Map map, Error&
...
 
-    entries.push_back(std::pair<Value, Value>(key, value));
+    if (descriptors.IsDoubleField(details)) {
+      double value;
+      if (index < 0) {
+        value = GetInObjectValue<double>(instance_size, index, err);
+      } else {
+        value = extra_properties.Get<double>(index, err);
+      }
+      // entries.push_back(std::pair<Value, Value>(key, value));
```
indutny: This needs to be either uncommented or deleted.

src/llscan.cc
```diff
@@ -397,11 +417,50 @@ bool FindReferencesCmd::DoExecute(SBDebugger d, char** cmd,
...
 
+  delete scanner;
+
   result.SetStatus(eReturnStatusSuccessFinishResult);
   return true;
 }
 
 
+char** FindReferencesCmd::ParseScanOptions(char** cmd, ScanType& type) {
+  static struct option opts[] = {{"value", no_argument, nullptr, 'v'},
```
indutny: Let's put a newline after first `{` and align everything nicely to the indent of 6?

src/llscan.cc
```diff
@@ -397,11 +417,50 @@ bool FindReferencesCmd::DoExecute(SBDebugger d, char** cmd,
...
+                                 {"name", no_argument, nullptr, 'n'},
+                                 {nullptr, 0, nullptr, 0}};
```
indutny: Space between `}` in `}}`.

src/llscan.cc
```diff
@@ -397,11 +417,50 @@ bool FindReferencesCmd::DoExecute(SBDebugger d, char** cmd,
...
+
+  optind = 0;
+  opterr = 1;
+
+  bool found_scan_type = false;
+
+  do {
+    int arg = getopt_long(argc, cmd - 1, "vn", opts, nullptr);
+    if (arg == -1) break;
+
```
indutny: Let's just put this here:

```c
if (found_scan_type) {
  type = ScanType::kBadOption;
  break;
}
```

src/llscan.cc
```diff
@@ -397,11 +417,50 @@ bool FindReferencesCmd::DoExecute(SBDebugger d, char** cmd,
...
+    switch (arg) {
+      case 'v':
+        type = found_scan_type ? ScanType::badoption : ScanType::fieldvalue;
```
indutny: No need in ternary if we'll go for the comment above.

src/llscan.cc
```diff
@@ -397,11 +417,50 @@ bool FindReferencesCmd::DoExecute(SBDebugger d, char** cmd,
...
+        found_scan_type = true;
+        break;
+      case 'n':
+        type = found_scan_type ? ScanType::badoption : ScanType::propertyname;
```
indutny: Ditto.

src/llscan.cc
```diff
@@ -397,11 +417,50 @@ bool FindReferencesCmd::DoExecute(SBDebugger d, char** cmd,
...
   }
 
+  delete scanner;
+
   result.SetStatus(eReturnStatusSuccessFinishResult);
   return true;
 }
 
 
+char** FindReferencesCmd::ParseScanOptions(char** cmd, ScanType& type) {
```
indutny: I'm not a huge fan of by-reference arguments. We use it for errors and some of lldb APIs, i.e. where it is inevitable, but let's use regular `*` in other places and pass arguments with `&` before them?

src/llscan.cc
hhellyer: Ok. It can drop straight out of the loop if it's found the option.
I've also dropped in changes to it matches the fix to the option parsing code under the other branch.
(The loop is slightly overkill until I add the option to limit the amount of output displayed but that should change shortly and in the meantime it looks like our other options parsing code.)

hhellyer:
On the subject of the options parsing code being excessive for only two mutually exclusive options there are at least two more things I want to add to this code:
- An option to limit the number of results returned.
- Another option to search by string value. (e.g find all the properties that refer to the string "Hello World!" with v8 findrefs -s "Hello World!")

(Suggestions for other ways of searching or other options are welcome!)

src/llscan.cc
```diff
@@ -381,13 +400,14 @@ bool FindReferencesCmd::DoExecute(SBDebugger d, char** cmd,
...
-        scanner.PrintRefs(result, js_obj, err);
+        scanner->PrintRefs(result, js_obj, err);
 
       } else if (type < v8->types()->kFirstNonstringType) {
         v8::String str(heap_object);
-        scanner.PrintRefs(result, str, err);
+        scanner->PrintRefs(result, str, err);
 
-      } else if (type < v8->types()->kJSTypedArrayType) {
+        // Should this be < or == ?!
```
indutny: I think it was just a copy-paste error.

src/llscan.cc
```diff
@@ -310,57 +311,75 @@ bool FindReferencesCmd::DoExecute(SBDebugger d, char** cmd,
...
-  v8::Value search_value(&llv8, value.GetValueAsSigned());
-  v8::Smi smi(search_value);
-  if (smi.Check()) {
-    result.SetError("Search value is an SMI.");
+  /* Ensure we have a map of objects.
+   * (Do this after we've checked the options to avoid
+   * a long pause before reporting an error.)
+   */
+  if (!llscan.ScanHeapForObjects(target, result)) {
+    result.SetStatus(eReturnStatusFailed);
```
indutny: Leaks `scanner`?

src/llscan.cc
```diff
@@ -474,6 +545,34 @@ void FindReferencesCmd::ReferenceScanner::PrintRefs(
...
+  v8::Value::InspectOptions inspect_options;
+  inspect_options.detailed = false;
+
+  // (Note: We skip array elements as they don't have names.)
+
+  // Walk all the properties in this object.
+  // We only create strings for the field names that match the search
+  // value.
+  std::vector<std::pair<v8::Value, v8::Value>> entries = js_obj.Entries(err);
+  if (err.Success()) {
```
indutny: `if (!err.Success()) return`?

hhellyer: That removes some indenting. (Not sure how it ended up like that.)
I've gone with `if (err.Fail())` to match the rest of the code in llscan.cc.

src/llscan.cc
```diff
@@ -310,57 +311,76 @@ bool FindReferencesCmd::DoExecute(SBDebugger d, char** cmd,
...
-  v8::Value search_value(&llv8, value.GetValueAsSigned());
-  v8::Smi smi(search_value);
-  if (smi.Check()) {
-    result.SetError("Search value is an SMI.");
+  /* Ensure we have a map of objects.
+   * (Do this after we've checked the options to avoid
+   * a long pause before reporting an error.)
+   */
+  if (!llscan.ScanHeapForObjects(target, result)) {
+    delete scanner;
```
indutny: Thinking more about it, should we just initialize scanner below? Is there any point of doing it before `ScanHeapForObjects` call?

hhellyer: The main reason for initialising it as part of the options handling was so that we could report any errors in the options before doing `ScanHeapForObjects`which may take quite a while. Otherwise the user sees a huge pause before being told they've entered a bad option.
It only really applies to the first run of the command (if it's the first one that needs to do a scan) as the results of the scan are cached.
I'll merge the change but I'm happy to go back and change it if you think it should. (Just raise an issue.)

PR #42, opened 2016-10-24
llnode.cc: Fix linux options parsing crash.

### LINKED ISSUE #41: v8 inspect crashes on Linux with invalid options
If you run v8 inspect -k 0x0 on Linux then llnode seg faults.
This is a problem with the handling of the command arguments in CommandBase::ParseInspectOptions.

hhellyer:
Fix up the arguments array to contain a first element rather than
subtract 1 from the lldb arguments and pretend they start at
args[1].
Linux prints args[0] in it's error message so it must point to a real
array element.
(Mac OSX appears to be hardcoded to use the program name in argv[0]
and ignores the first element of the arguments array.)

Fixes #41

### DIFF
--- a/src/llnode.cc
+++ b/src/llnode.cc
@@ -33,7 +33,7 @@ char** CommandBase::ParseInspectOptions(char** cmd,
   // for getopt_long to use in its error messages.
   char name[] = "llnode";
   args[0] = name;
-  for (int i = 0; i < argc; i++) args[i+1] = cmd[i];
+  for (int i = 0; i < argc - 1; i++) args[i + 1] = cmd[i];
 
   // Reset getopts.
   optind = 1;
@@ -61,7 +61,7 @@ char** CommandBase::ParseInspectOptions(char** cmd,
   } while (true);
 
   // Use the original cmd array for our return value.
-  return cmd + (optind - 1);
+  return &cmd[optind - 1];
 }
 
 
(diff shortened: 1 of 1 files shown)


src/llnode.cc
```diff
@@ -24,13 +24,22 @@ char** CommandBase::ParseInspectOptions(char** cmd,
...
   for (char** p = cmd; p != nullptr && *p != nullptr; p++) argc++;
 
-  optind = 0;
+  char* args[argc];
+
+  // Make this look like a command line, we need a valid element at index 0
+  // for getopt_long to use in its error messages.
+  char name[] = "llnode";
+  args[0] = name;
+  for (int i = 0; i < argc; i++) args[i+1] = cmd[i];
```
indutny: Off by one write here? Also style `i + 1`.

src/llnode.cc
```diff
@@ -51,7 +60,8 @@ char** CommandBase::ParseInspectOptions(char** cmd,
     }
   } while (true);
 
-  return cmd + optind - 1;
+  // Use the original cmd array for our return value.
+  return cmd + (optind - 1);
```
indutny: Good catch!

indutny: Shoudl we use `&cmd[optind - 1]` instead?

indutny:
Thank you!

PR #46, opened 2016-11-12
llscan: process chunks in blocks for speed boost

indutny:
This improves time of `ScanMemoryRanges` for 1.3gb core file by 2x times
(from 103s to 52s).

cc @hhellyer @yjhjstz

### DIFF
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -623,7 +623,7 @@ uint64_t FindJSObjectsVisitor::Visit(uint64_t location, uint64_t word) {
       map_info.type_name = heap_object.GetTypeName(&inspect_options, err);
 
     // Cache result
-    map_cache_.insert(std::pair<int64_t, MapCacheEntry>(map.raw(), map_info));
+    map_cache_.emplace(map.raw(), map_info);
 
     if (err.Fail()) return address_byte_size_;
   } else {
@@ -637,8 +637,7 @@ uint64_t FindJSObjectsVisitor::Visit(uint64_t location, uint64_t word) {
     TypeRecord* t = new TypeRecord(map_info.type_name);
 
     t->AddInstance(word, map.InstanceSize(err));
-    mapstoinstances_.insert(
-        std::pair<std::string, TypeRecord*>(map_info.type_name, t));
+    mapstoinstances_.emplace(map_info.type_name, t);
 
   } else {
     /* Update an existing instance, if we haven't seen this instance before. */
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -586,15 +586,7 @@ FindJSObjectsVisitor::FindJSObjectsVisitor(SBTarget& target,
 
 
 /* Visit every address, a bit brute force but it works. */
-uint64_t FindJSObjectsVisitor::Visit(uint64_t location, uint64_t available) {
-  lldb::SBError error;
-
-  // Test if the map points to a real map.
-  // Try to create an object out of it.
-
-  uint64_t word = target_.GetProcess().ReadUnsignedFromMemory(
-      location, address_byte_size_, error);
-
+uint64_t FindJSObjectsVisitor::Visit(uint64_t location, uint64_t word) {
   v8::Value v8_value(&llv8, word);
 
   v8::Error err;
@@ -745,6 +737,12 @@ void LLScan::ScanMemoryRanges(FindJSObjectsVisitor& v) {
 #ifndef LLDB_SBMemoryRegionInfoList_h_
   MemoryRange* head = ranges_;
 
+  const uint64_t addr_size = process_.GetAddressByteSize();
+
+  // Pages are usually around 1mb, so this should more than enough
+  const uint64_t block_size = 1024 * 1024 * addr_size;
+  unsigned char* block = new unsigned char[block_size];
+
   while (head != nullptr && !done) {
     uint64_t address = head->start_;
     uint64_t len = head->length_;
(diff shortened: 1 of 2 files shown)


src/llscan.cc
```diff
@@ -768,16 +766,46 @@ void LLScan::ScanMemoryRanges(FindJSObjectsVisitor& v) {
...
+    uint64_t address_end = address + len;
+
+    // Load data in blocks to speed up whole process
+    for (auto searchAddress = address; searchAddress < address_end;
+         searchAddress += block_size) {
+      size_t loaded = std::min(address_end - searchAddress, block_size);
+      process_.ReadMemory(searchAddress, block, loaded, sberr);
+      if (sberr.Fail()) {
+        // TODO(indutny): add error information
+        fprintf(stderr, "fail\n");
```
indutny: Yikes, I'll remove this.

indutny:
Added one more commit that makes `findrefs` work on my core dump.

indutny:
One more patch and the scanning time now is 22 seconds.

src/llscan.cc
```diff
@@ -586,69 +590,59 @@ FindJSObjectsVisitor::FindJSObjectsVisitor(SBTarget& target,
...
-  inspect_options.detailed = false;
-  inspect_options.print_map = false;
-  inspect_options.print_source = false;
-  inspect_options.string_length = 0;
+    // On success load type name
+    if (map_info.is_histogram)
+      map_info.type_name = heap_object.GetTypeName(&inspect_options, err);
+
+    // Cache result
+    map_cache_.insert(std::pair<int64_t, MapCacheEntry>(map.raw(), map_info));
```
yjhjstz: can we use C++11 `map.emplace` ?

src/llscan.cc
```diff
@@ -745,6 +740,12 @@ void LLScan::ScanMemoryRanges(FindJSObjectsVisitor& v) {
 #ifndef LLDB_SBMemoryRegionInfoList_h_
   MemoryRange* head = ranges_;
 
+  const uint64_t addr_size = process_.GetAddressByteSize();
+
+  // Pages are usually around 1mb, so this should more than enough
+  const uint64_t block_size = 1024 * 1024 * addr_size;
```
yjhjstz: addr_size << 20 ?

src/llscan.cc
```diff
@@ -768,16 +769,45 @@ void LLScan::ScanMemoryRanges(FindJSObjectsVisitor& v) {
...
+        else
+          break;
+
+        increment = v.Visit(j + searchAddress, value);
+        if (increment == 0) break;
+
+        j += static_cast<size_t>(increment);
+      }
+
       if (increment == 0) {
```
yjhjstz: should here skip search address? do not break , but continue?

indutny: I believe @hhellyer meant end of iteration with this. It doesn't seem to be used right now, though.

src/llscan.cc
indutny: This is a bit cleaner, IMO. Isn't it?

src/llscan.cc
indutny: Fixed!

yjhjstz:
👍

indutny:
Thank you, @yjhjstz !

@hhellyer your feedback is highly welcome on this too!

src/llscan.cc
hhellyer: block_size is only defined inside #ifndef LLDB_SBMemoryRegionInfoList_h_. If you have the new API then block_size is undefined and llscan.cc doesn't compile.
Ubuntu 16.10 now has an lldb-3.9 package with the new API.
(The path that uses the SBMemoryRegionInfo API also speeds things up by only scanning writable memory regions.)

hhellyer:
Looks good, I'm just going through the diffs to do a proper review.
I did a couple of comparison runs with a simple core from --abort_on_uncaught_exception and one from an out of memory where I just filled the heap until Node crashed:

Minimal core dump (small Node.js heap):
```
llnode master:						74s
llnode feature/speed-up:			14s
llnode master, lldb 3.9:				20s
llnode feature/speed-up, lldb 3.9:	 3s
```

That shows lldb 3.9 as faster than the default version, probably because most of the core file is read only memory. The speed up helps quite a lot there too.

OOM Core dump:
```
llnode master, lldb 3.9:				24m16.170s
llnode feature/speed-up, lldb3.9:		 3m30.101s
```

With a heap dump where most of the core is writable memory containing JavaScript objects the improvement from this PR is massive. (It failed to do the other two, otool failed to open the 2.7 Gb core dump so I couldn't actually generate the ranges file. I'll have to investigate that.)

src/llscan.cc
```diff
@@ -586,69 +590,58 @@ FindJSObjectsVisitor::FindJSObjectsVisitor(SBTarget& target,
...
-  if (err.Fail()) {
-    return address_byte_size_;
-  }
+  MapCacheEntry map_info;
+  if (map_cache_.count(map.raw()) == 0) {
+    v8::Value::InspectOptions inspect_options;
+    inspect_options.detailed = false;
+    inspect_options.print_map = false;
+    inspect_options.print_source = false;
+    inspect_options.string_length = 0;
```
hhellyer: There's nothing wrong with this change except that it reminds me I ought to remove the inspect_options parameter from GetTypeName. It shouldn't need them as it only ever returns a type name (not the source or the map and it doesn't truncate it if it's a long string).

indutny: Something to be fixed later, right?

indutny:
@hhellyer should I merge it now? 😉

hhellyer:
LGTM!

indutny:
Landed in many commits, thank you everyone!

Issue #50, opened 2016-11-15
Internal fields


indutny:
It would be lovely if we could inspect C++ from object's internal fields. This will require some additional constants to be exported.

hhellyer:
Agreed, we'd quite like to be able to get things like heap roots or the start and end of the various v8 heaps.
Being able to find out what callbacks are sat on the libuv event loop would also be good - though that might be part of having the heap roots since they'd be keeping objects alive.

indutny:
While this is cool too, I actually meant: http://bespin.cz/~ondras/html/classv8_1_1Object.html#a435f68bb7ef0f64dd522c5c910682448

PR #51, opened 2016-11-18
JSAPIObject and Internal Fields

indutny:
see #50 

cc @hhellyer @yjhjstz

### DIFF
--- a/src/llv8-constants.cc
+++ b/src/llv8-constants.cc
@@ -124,15 +124,34 @@ void Common::Load() {
   kPointerSize = 1 << LoadConstant("PointerSizeLog2");
   kVersionMajor = LoadRawConstant("v8::internal::Version::major_");
   kVersionMinor = LoadRawConstant("v8::internal::Version::minor_");
+  kVersionPatch = LoadRawConstant("v8::internal::Version::patch_");
 }
 
 
-bool Common::CheckVersion(int64_t major, int64_t minor) {
+bool Common::CheckHighestVersion(int64_t major, int64_t minor, int64_t patch) {
   Load();
 
-  if (major > kVersionMajor) return false;
-  if (minor > kVersionMinor) return false;
+  if (kVersionMajor < major) return true;
+  if (kVersionMajor > major) return false;
 
+  if (kVersionMinor < minor) return true;
+  if (kVersionMinor > minor) return false;
+
+  if (kVersionPatch > patch) return false;
+  return true;
+}
+
+
+bool Common::CheckLowestVersion(int64_t major, int64_t minor, int64_t patch) {
+  Load();
+
+  if (kVersionMajor > major) return true;
+  if (kVersionMajor < major) return false;
+
+  if (kVersionMinor > minor) return true;
+  if (kVersionMinor < minor) return false;
+
+  if (kVersionPatch < patch) return false;
   return true;
 }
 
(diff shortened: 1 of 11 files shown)


indutny:
I'll write a test in a bit.

indutny:
Pushed a test.

indutny:
@nodejs/jenkins-admins we really need to have a CI coverage for this repo. Please help us!

src/llv8-constants.cc
```diff
@@ -472,6 +500,8 @@ void Types::Load() {
       LoadConstant("type_JSGlobalObject__JS_GLOBAL_OBJECT_TYPE");
   kOddballType = LoadConstant("type_Oddball__ODDBALL_TYPE");
   kJSObjectType = LoadConstant("type_JSObject__JS_OBJECT_TYPE");
+  kJSAPIObjectType = LoadConstant("APIObjectType");
```
indutny: I'm going to send a CL to v8 team for these constants.

indutny:
V8 CL: https://codereview.chromium.org/2514063002/

src/llv8.cc
```diff
@@ -1179,13 +1179,52 @@ std::string JSObject::Inspect(InspectOptions* options, Error& err) {
...
+  int64_t instance_size = map.InstanceSize(err);
+
+  // kVariableSizeSentinel == 0
+  // TODO(indutny): post-mortem constant for this?
+  if (err.Fail() || instance_size == 0) return std::string();
+
+  std::string res;
+  for (int64_t off = v8()->js_object()->kInternalFieldsOffset;
+       off < instance_size; off += v8()->common()->kPointerSize) {
+    int64_t field = LoadField(off, err);
```
yjhjstz: seem we can't inspect internal fieldd in a recursive way.

indutny: Internal field is a pointer to C++ object. We can't really do it.

src/llv8-constants.cc
indutny: Sent.

indutny:
Does it LGTY @yjhjstz (and @hhellyer )?

rvagg:
@indutny whatchu need that travis isn't providing? GIve us a baseline of platform requirements and we'll see what we can hook up.

indutny:
@rvagg I need just travis, actually, but we can't enable it without giving access to all private repos in this org, AFAIK.

indutny:
@yjhjstz @hhellyer pushed tests for llscan here too. PTAL.

indutny:
CI: https://ci.nodejs.org/job/llnode/9/

indutny:
Internal field inspection doesn't work that well atm, as it account in-object properties too. Will look into fixing it.

indutny:
Added fix to account for in-object properties.

hhellyer:
Changes look good but it is worth adding the code that uses the extra constants from gen-postmortem-metadata.py before that patch has made it into v8 (and then into node)?
Would it be better to split those out from the scan tests and put the tests in under a separate PR?

indutny:
@hhellyer that V8 CL is about to land (just got LGTM), and I think it is quite usual practice in this repo to introduce them at the same time if fallback is possible.

indutny:
@yjhjstz does it LGTY?

indutny:
@hhellyer CL landed, I don't really feel like splitting PR at this point. Is it ok?

indutny:
Landed in ad45a96, 4459f90, and b7b0306. Thank you!

indutny:
We can always revert these changes if anything will go wrong.

PR #54, opened 2016-11-21
llv8: function's inferred name may be undefined

indutny:
Discovered while testing on Ubuntu.

cc @hhellyer @yjhjstz


yjhjstz:
👍

indutny:
Landed in 228481a, thank you!

PR #57, opened 2016-11-24
llnode: properly reset `getopt` on linuxes

indutny:
When `getopt` is called it starts seeking from the last char it has
seen. Unfortunately, that last char may be part of the string that was
already deallocated.

Set `optind` to `0` to reset the state properly on Linux.

See: https://github.com/nodejs/llnode/pull/35#issuecomment-262854793

cc @hhellyer @yjhjstz

### DIFF
--- a/src/llnode.cc
+++ b/src/llnode.cc
@@ -36,7 +36,7 @@ char** CommandBase::ParseInspectOptions(char** cmd,
   for (int i = 0; i < argc - 1; i++) args[i + 1] = cmd[i];
 
   // Reset getopts.
-  optind = 1;
+  optind = 0;
   opterr = 1;
   do {
     int arg = getopt_long(argc, args, "Fms", opts, nullptr);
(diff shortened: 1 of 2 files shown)


indutny:
CI: https://ci.nodejs.org/job/llnode/14/

indutny:
This still doesn't fix this, but at least it fixes more important problem:

```
(lldb) 8 bt
error: '8' is not a valid command.
error: Unrecognized command '8'.
vnot ok 1 test timed out after 15000ms
```

hhellyer:
I think there's differences between Linux and Mac implementations of getopt. I can't reliably reproduce the problem on my Ubuntu VM but "man 3 getopt" says in it's notes section that optind should be reset to 0 in some circumstances to force some internal reinitialisation.

I'm not sure if we're hitting the conditions it specifies but it might be worth trying if you can reproduce the problem easily.

indutny:
@hhellyer gosh, I swear I tried it before... and it wasn't working, but it works now. Furthermore I can see in getopt source that it should reset it state. Will change PR to do that.

indutny:
CI: https://ci.nodejs.org/job/llnode/15/

indutny:
@hhellyer changed, PTAL

hhellyer:
LGTM - Or at least doesn't seem to introduce any new problems on Linux or Mac!

indutny:
Landed in a093e2a, thank you!

Issue #53, opened 2016-11-21
npm i llnode


Fishrock123:
I think this should be possible to install on a global install via npm scripts, @indutny is there any reason you didn't go this route?

indutny:
I don't mind if someone will make it work :wink:

hhellyer:
It would definitely be nice to have a simpler way to install llnode, especially on Linux.
I'm not 100% sure on using npm to install things that aren't going to be node code but there's also the possibility of adding a javascript API to llnode in which case it would make more sense.

It could write a plugin load command to ~/.lldbinit instead of needing root access to install the plugin in lldb's plugin directory. There will probably be issues around using the right version of lldb, on ubuntu it's possible to have multiple versions installed as lldb-3.8, lldb-3.9.

I'm happy to take a look at doing this. Should we reserve the name llnode on npmjs.org?

indutny:
Sure, go ahead!

Fishrock123:
Two things: make sure you register with a version that `llnode` has not used, like 0.0.0. Also, we should publish some basic install instructions there in the mean time.

hhellyer:
I've grabbed the llnode npm name. I've added @indutny as a collaborator. (Assuming that isn't someone else with the same picture on npmjs.org.)
I assume the npm installer should work for Mac and Linux (as a minimum) and I'll probably look at what the brew installer is doing.

indutny:
(Yep, that's me)

hhellyer:
@indutny Making progress on this but I've got a question, why did you use llnode.gyp instead of binding.gyp?

The background to this is that for some reason npm doesn't copy llnode.gyp unless you specify it explicitly (and then have to specify everything to copy). I think there's something special about <packagename>.gyp but I don't know what. Also you can't stop npm copying binding.gyp.
I may end up changing the build process slightly, I'm also trying to figure out if it's possibly to use the gyp packaged with npm rather than cloning it during the build process so the only thing that may need cloning is the lldb headers.

hhellyer:
I've pushed a branch https://github.com/hhellyer/llnode/tree/npm_prototype which you are welcome to try however it's not final, part of the reason for pushing to github is to make it easy to test it on more machines.

indutny:
@hhellyer splendid! Will take a look.

btw, I forgot to mention it, but I have solved similar problem in bud: https://github.com/indutny/bud

hhellyer:
[...]

PR #60, opened 2016-12-06
package.json: Make llnode installable via npm

hhellyer:
This updates package.json and adds scripts to allow llnode to be
built via "npm install llnode"
It doesn't create a Node.js module you can use via require('llnode').
(This may change in the future depending on project direction.)

The npm name llnode has been reserved here:
https://www.npmjs.com/package/llnode
and Fedor also has access to that as discussed under issue #53.

This PR does leave the npm version number at 0.0.0, this should
be updated once the install process has been reviewed and at
that point the llnode package should be npm publish'd.


package.json
```diff
@@ -1,18 +1,31 @@
...
   "description": "llnode test suite",
   "main": "no-entry-sorry.js",
   "directories": {
     "test": "test"
   },
+  "//": "We need to use the install scripts as llnode must be built on the machine it's deployed on.",
+  "//": "(Blame C++)",
   "scripts": {
+    "preinstall" : "node scripts/configure.js",
+    "install" : "./gyp_llnode && make -C out/",
```
richardlau: I'm new to llnode, but is there a reason this couldn't be built with a `binding.gyp` which would automatically cause `npm` to invoke its bundled `node-gyp`?

hhellyer: I was trying to leave the build process the same for cloning from git and installing via npm really.

yjhjstz:
cool!

package.json
rnchamberlain: I wonder if we could support building the plugin in the 'clone from git' case using 'node-gyp rebuild' - that seems to be popular for node.js natives. Best ask @indutny

hhellyer:
I've updated configure.js to do a better job of finding the lldb executable on Linux.
It also now allows the user to force the use of a particular build of lldb by letting the user specify --lldb_exe on the command line. So on Ubuntu to force the user of lldb-3.9 you can do:

```npm install --lldb_exe=`which lldb-3.9` llnode```

**Comment by hhellyer** [comment]
@indutny @Fishrock123 - I'll be out from the end of this week to 3rd January. I'm happy to make changes this week if you have any comments. If you don't have time this week I don't mind if you make changes on my branch while I'm out, otherwise we can fix it up in the new year.

File: package.json (hunk 1 above)
Comment (gibfahn) [inline package.json]: >I was trying to leave the build process the same for cloning from git and installing via npm really.

Is there a reason you couldn't just `git clone https://github.com/nodejs/llnode.git && npm install` (i.e. use `node-gyp rebuild` for both)?

**Comment by hhellyer** [comment]
I've put this branch up on npmjs.org as version 0.0.1 so that it can get some testing and feedback when used as an actual npm.
https://www.npmjs.com/package/llnode

File: scripts/cleanup.js [hunk 2]
```diff
@@ -0,0 +1,19 @@
+var os = require('os');
```
indutny: Let's use ES6 😉 :

```
'use strict';
const os = ...
```

scripts/cleanup.js
```diff
@@ -0,0 +1,19 @@
...
+var child_process = require('child_process');
+
+var cwd = process.cwd();
+var osName = os.type();
+var libExt = "so";
```
indutny: `'` instead of `"`

scripts/cleanup.js
```diff
@@ -0,0 +1,19 @@
...
+
+if (osName === 'Darwin') {
+  libExt = "dylib";
+}
+
+var llnodeLib = "llnode." + libExt;
+
+// Move the library somewhere easy to remember.
+console.log('Copying ' + cwd + '/out/Release/' + llnodeLib + ' to ' + cwd + '/' + llnodeLib);
```
indutny: Template strings:

```js
`Copying ${cwd} ...`
```

scripts/configure.js
```diff
@@ -0,0 +1,192 @@
...
+
+  if(lldbVersion === undefined) {
+    console.log('Unable to locate lldb binary. llnode installation failed.');
+    process.exit(1);
+  }
+
+  lldbHeadersBranch = lldbReleases[lldbVersion];
+  lldbIncludeDir = 'lldb-' + lldbVersion;
+
+} else if ( osName === 'Linux') {
```
indutny: Style: `if (osName === 'Linux') {`

scripts/configure.js
```diff
@@ -0,0 +1,192 @@
...
+// Need to determine:
+// - What level of lldb we are running.
+// - If we need the headers. (Linux may have them installed)
+var lldbExe = 'lldb';
+
+if (osName === 'Darwin') {
+
+  lldbVersion = getDarwinRelease();
+
+  if(lldbVersion === undefined) {
```
indutny: `if (`

indutny: I think we may want to use `eslint` here.

scripts/configure.js
```diff
@@ -0,0 +1,192 @@
...
+}
+
+// Find the 'best' lldb to use. Either:
+// - the one specified by the user using npm --llnode_exe=... install llnode
+// - the default lldb executable
+// - the higest known lldb version
+function getLldbExecutable() {
+  var lldbExe = process.env.npm_config_lldb_exe;
+  if (lldbExe === undefined) {
+    var lldbExeNames = ['lldb', 'lldb-4.0', 'lldb-3.9', 'lldb-3.8', 'lldb-3.7', 'lldb-3.6']
```
indutny: Missing semicolon

rnchamberlain:
FYI, tried this out on Ubuntu, lldb 3.9, with a core dump from Node.js v7:
```
> npm install llnode
> lldb node -c core
(lldb) plugin load node_modules/llnode/llnode.so
(lldb) jsstack
```
Works well, this is going to be nice

package.json
Fishrock123: ^ That's what I was thinking?

hhellyer: I think you'd have to ask @indutny why he originally didn't use node-gyp. I would add though that adding that to the build process adds a dependency on having node. When you build llnode from github there's not actually any dependancy on node.

There's a major dependancy on having the right lldb headers and you do need gyp in some form but weirdly for a node debugging tool there's no dependancy on node itself. llnode is built against a specific version of lldb but will work against a wide range of node versions - including being able to open Linux cores with lldb/llnode on Mac. If we did move to node-gyp we'd want to avoid introducing a link to the version of node that was running node-gyp.

For my part I wanted to focus on making the changes necessary to configure and build via npm to make it easier for people to get hold of rather than re-writing the build process. Moving to node-gyp isn't necessarily a bad idea but it should probably be raised as a separate issue.

hhellyer:
I copied the .eslintrc from a node build and ran that against the scripts. They should be formatted correctly now.

package.json
indutny: I'm fine with it, but let's handle it separately.

hhellyer:
@indutny I'll put the package.json level back to 1.4.0, do a merge and then publish a v1.4.0 package to npm. Is that reasonable? (We might need to document the publish process the contribution docs now too.)

indutny:
@hhellyer sounds reasonable to me.

hhellyer:
I've published to https://www.npmjs.com/package/llnode - it took an hour or so to update the npmjs.org page last time so that still shows 0.0.1 but doing `npm install llnode` now installs version 1.4.1.

indutny:
Fantastic! Thank you!

bnoordhuis:
@hhellyer Why did you rename/symlink llnode.gyp to llnode.gyp.json?  From looking at 2325ae3 it's quite clearly a deliberate move but for the life of me, I can't figure out why.

hhellyer:
@bnoordhuis - Well I'm quite prepared to be told I was being an idiot but no matter what I did I couldn't figure out why when I did npm install <path_to_folder> (during development) the npm never copied files ending in .gyp. So I renamed it llnode.gyp.json. (I've just re-tested it and calling it llnode.gyp or build.gyp both still cause it not to be copied.)

It might be an obscure npm feature or just something really silly I was doing but in the end renaming it seemed enough to solve my problem. There might be some weirdness around having .gyp files other than binding.gyp in your npm, I just added one to something else to verify the copying problem and it's existence alone broke the build.

I think I just put the symlink in so I didn't break anyone else who was referring to the file. (I was trying to make the minimum set of changes.)

richardlau:
I don't have the reference to hand but I believe npm has (had?) the surprising behaviour of renaming all `*.gyp` files to `binding.gyp`.

hhellyer:
@richardlau - That would actually fit what I saw. (And explain why now if I change binding.gyp in the npm I'm working on to iamafish.gyp it still builds!)

richardlau:
https://github.com/npm/npm/issues/8411 and https://github.com/npm/fstream-npm/pull/18

PR #62, opened 2016-12-15
llnode.cc: Only mention ranges file if it is needed.

hhellyer:
Since we are starting to see people using lldb 3.9 in the field (at least on Linux) this just makes sure they don't see confusing help messages that don't apply since 3.9 has the SBMemoryRegionInfoList API available.

### DIFF
--- a/src/llnode.cc
+++ b/src/llnode.cc
@@ -319,7 +319,7 @@ bool PluginInitialize(SBDebugger d) {
                 "debugged.\n"
                 "There are scripts for generating this file on Linux and Mac "
                 "in the scripts directory of the llnode repository."
-#endif // LLDB_SBMemoryRegionInfoList_h_
+#endif  // LLDB_SBMemoryRegionInfoList_h_
                 );
 
   interpreter.AddCommand("findjsobjects", new llnode::FindObjectsCmd(),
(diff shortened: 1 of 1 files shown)


rnchamberlain:
LGTM. 

Is there something we can do in the project README as well> Currently it says build against lldb 3.8, which would still need the ranges file of course.

src/llnode.cc
```diff
@@ -313,11 +313,14 @@ bool PluginInitialize(SBDebugger d) {
...
                 "List all object types and instance counts grouped by map and "
                 "sorted by instance count.\n"
+#ifndef LLDB_SBMemoryRegionInfoList_h_
                 "Requires `LLNODE_RANGESFILE` environment variable to be set "
                 "to a file containing memory ranges for the core file being "
                 "debugged.\n"
                 "There are scripts for generating this file on Linux and Mac "
-                "in the scripts directory of the llnode repository.");
+                "in the scripts directory of the llnode repository."
+#endif // LLDB_SBMemoryRegionInfoList_h_
```
indutny: Minor nit: two spaces between `#endif` and `//`

hhellyer: Done. (Hmm, github doesn't have a tick reaction to mark review comments fixed.)

indutny:
Thank you!

indutny:
@hhellyer may I ask you to put metadata into commit as in: https://github.com/nodejs/llnode/commit/3fd5d0b1550eef3d5dec5a5a6f234288666a94df in the future? (`PR-URL`, `Reviewed-By`)

hhellyer:
Sure, should we update CONTRIBUTING.md with that as well?

indutny:
Yeah, that would be lovely!

PR #67, opened 2017-02-28
package.json: update project description

rnchamberlain:
The description attribute in the package.json file appears as headline on https://www.npmjs.org/package/llnode. The current description 'llnode test suite' is a little confusing, as the module is an LLDB plugin not a test suite. 

This PR fixes it to match the introduction in the README.


indutny:
Landed in 849b5b2, thank you!

Issue #68, opened 2017-02-28
otool2segments does not work anymore


indutny:
```
$ ~/Code/indutny/llnode/scripts/otool2segments.py ./core
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/objdump: './core': Invalid argument.
fatal error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/otool: internal objdump command failed
```

cc @hhellyer

rnchamberlain:
Do we still need the memory segments tool? The better fix in lldb is in v3.9 and later. Maybe that's not available on Mac?

indutny:
Its not available on Mac, at least out of the box.

hhellyer:
I ran a quick sniff test with a random version of node and it looks ok initially. It's definitely a Mac core dump? (All mine end up in /cores named core.<pid> I'm not sure how to get them called "core" on a Mac.)
I'll try and re-create if you give me some more details:
- Which version of Node created the core? 32 or 64 bit?
- Which version of otool? (otool --version says (Apple LLVM version 8.0.0 (clang-800.0.42.1)) for me.)
- Can lldb open the core dump? (It could just be a broken core.)

indutny:
This is definitely a core, I just moved it to `/tmp/` for unrelated reasons:

```
64 bit node.js v7.6.0
```

```
llvm-otool(1): Apple Inc. version cctools-895
Apple LLVM version 8.0.0 (clang-800.0.42.1)
```

```
$ file /tmp/core
/tmp/core: Mach-O 64-bit core x86_64
```

indutny:
I believe that the latest `otool` internally calls `objdump` now. Here is `objdump`:

```
Apple LLVM version 8.0.0 (clang-800.0.42.1)
  Optimized build.
  Default target: x86_64-apple-darwin16.4.0
  Host CPU: haswell
```

hhellyer:
That's the same level of otool as me.
How big is it? I have a recollection I managed to break otool with a really massive core dump.
LLDB might be able to still open it.

indutny:
Hm... It is around 3gb: `2876063744`. Is it too big?

indutny:
`lldb` works with it indeed.

hhellyer:
I just tried a test program that filled the heap and crashed. The core dump was 2889756672 and otool couldn't open it.

[...]

PR #69, opened 2017-03-01
scripts: add JS-only script for large cores on Mac

indutny:
Fix: #68

### DIFF
--- a/scripts/macho2segments.js
+++ b/scripts/macho2segments.js
@@ -1,0 +1,78 @@
+#!/usr/bin/env node
+'use strict';
+
+const assert = require('assert');
+const fs = require('fs');
+const Buffer = require('buffer').Buffer;
+
+const ALIGN_64 = 8;
+
+const HEADER_SIZE = 4 * 7;
+const BODY_OFF_64 = 4 * 8;
+const MAGIC_LE_32 = 0xfeedface;
+const MAGIC_LE_64 = 0xfeedfacf;
+
+const CMD_HEADER_SIZE = 8;
+const LC_SEGMENT_64 = 0x19;
+const LC_SEGMENT_64_SIZE = 4 * 2 + 16 + 8 * 4 + 4 * 4;
+const VM_PROT_WRITE = 0x2;
+
+if (process.argv.length < 3) {
+  console.error('Usage: %s <core_file>', process.argv[1]);
+  process.exit(1);
+  return;
+}
+
+const fd = fs.openSync(process.argv[2], 'r');
+
+function read(start, end) {
+  const data = Buffer.alloc(end - start);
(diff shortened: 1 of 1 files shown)


indutny:
This is a very limited script as it supports only 64-bit Little Endian platforms. We can potentially expand it later if Apple won't update lldb earlier :wink:

scripts/macho2segments.js
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env node
+'use strict';
+
+const assert = require('assert');
+const fs = require('fs');
+const Buffer = require('buffer').Buffer;
```
cjihrig: We don't really need to require this, right?

scripts/macho2segments.js
```diff
@@ -0,0 +1,78 @@
...
+  return;
+}
+
+const fd = fs.openSync(process.argv[2], 'r');
+
+function read(start, end) {
+  const data = Buffer.alloc(end - start);
+  const bytesRead = fs.readSync(fd, data, 0, data.length, start);
+
+  assert.equal(data.length, bytesRead, 'Read less than expected');
```
cjihrig: Any reason not to use `strictEqual()` throughout this file?

scripts/macho2segments.js
indutny: That's how I always roll :) Maybe one day we will be able to remove global `Buffer`.

scripts/macho2segments.js
indutny: Ack.

indutny:
@cjihrig your feedback is very welcome!

scripts/macho2segments.js
cjihrig: Very wishful thinking :-)

hhellyer:
@indutny - I compared it to the output from the otool script on a smaller dump. It gets the section sizes wrong due to some the string concatenation in parseUInt64LE.

I switched to
```
function parseUInt64LE(buf, off) {
  var low = buf.readUInt32LE(off).toString(16);
  var high = buf.readUInt32LE(off + 4).toString(16);
  while (high.length < 8)
    high = '0' + high;
  while (low.length < 8)
    low = '0' + low;
  return high+low;
}
```
and got matching output.

That needs fixing but I'm pretty impressed with how quickly you parsed it otherwise!

It also picks up a couple of extra ranges that otool doesn't and misses off the leading 0x from the hex output. The ranges are probably fine - otool probably has a reason to hide them but it probably doesn't affect us and if C++ std::hex doesn't mind missing the leading 0x neither do I.

indutny:
@hhellyer fixed mentioned nits, PTAL.

indutny:
Landed in 1a27c72, thank you!

PR #70, opened 2017-03-01
Fix bug in v8 findrefs command, reference search fails

rnchamberlain:
Line 353 in llscan.c creates a `ReferenceScanner`, passing in a `v8::Value `for the search parameter. The parameter on the `ReferenceScanner `constructor is defined and hence passed as a C++ reference, but the scope of the `v8::Value` passed in is block local, see line 346 in llscan.c. So the search value is a wild reference when used later.

Fix is to change the definition of the `ReferenceScanner` constructor to pass the parameter by value.

### DIFF
--- a/src/llscan.h
+++ b/src/llscan.h
@@ -56,7 +56,7 @@ class FindReferencesCmd : public CommandBase {
 
   class ReferenceScanner : public ObjectScanner {
    public:
-    ReferenceScanner(v8::Value& search_value) : search_value_(search_value) {}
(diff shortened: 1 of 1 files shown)


rnchamberlain:
And he'll want to buy me a pint

hhellyer:
LGTM but I'm sure it takes more than deleting 2 characters to be owed a pint.

indutny:
Landed in 89c3094a9d3329f7f2daf4985f0836ebe6f1d23b, thank you!

PR #72, opened 2017-03-03
Fix and docs for lldb system plugin directory

### LINKED ISSUE #71: Linux lldb system plugin directory has changed
See https://reviews.llvm.org/D19516. The lldb system plugins directory for linux is now` /usr/lib/lldb/plugins`. I think this happened in lldb v3.9. So it no longer finds the llnode plugin if it is installed in `/usr/lib/lldb` (which is where the current llnode makefile install target puts it). 

They say that earlier lldb versions do a recursive scan from /usr/lib/lldb, so lldb v3.8 should still find our plugin if we install it in` /usr/lib/lldb/plugins`

rnchamberlain:
Fix makefile to install llnode as system plugin for linux in /usr/lib/lldb/plugins not /usr/lib/lldb. Also add README docs for usage of lldb system plugin directory vs `plugin load`.

Fixes #71 

We could have the npm install also do the copy to the system plugin directory, but on Linux that would need root/sudo. Maybe it could be an extra option on the npm install. Anyway, I think it's worth having this documented in the README as it's not obvious what's going on.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,11 +10,11 @@ uninstall-osx:
 	rm ~/Library/Application\ Support/LLDB/PlugIns/llnode.dylib
 
 install-linux:
-	mkdir -p /usr/lib/lldb
-	cp -rf ./out/Release/lib.target/llnode.so /usr/lib/lldb/
+	mkdir -p /usr/lib/lldb/plugins
+	cp -rf ./out/Release/lib.target/llnode.so /usr/lib/lldb/plugins
 
 uninstall-linux:
-	rm /usr/lib/lldb/llnode.so
+	rm /usr/lib/lldb/plugins/llnode.so
 
 format:
 	clang-format -i src/*
--- a/README.md
+++ b/README.md
@@ -75,10 +75,38 @@ make -C out/ -j9
 sudo make install-linux
 ```
 
-### Usage
+## Usage
+
+The plugin call be loaded in LLDB using the `plugin load` command, or
+installed in the LLDB system plugin directory. LLDB will then load the
+plugin automatically on start-up. 
+
+### OS X
 
 ```
 lldb
+
+(lldb) plugin load ./node-modules/llnode/llnode.dylib
+```
+
+To install the plugin in the LLDB system plugin directory, use the
+`make install-osx` build step above, or if installing
+with npm copy `node_modules/llnode/llnode.dylib` to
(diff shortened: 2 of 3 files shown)


gibfahn:
Maybe have an [npm run script](https://github.com/nodejs/llnode/blob/master/package.json#L11-L16) called `npm run system-install` or something, that does the copy to the llnode system plugin directory. You could then have the [postinstall script](https://github.com/nodejs/llnode/blob/master/scripts/cleanup.js) echo:

```
To load automatically on lldb startup, run `sudo npm run system-install` to copy to the lldb system plugin directory.
```

So at the end people installing it would see the option.

Documenting it in the README is definitely a good idea as well.

rnchamberlain:
Maybe a bit over-the-top to have the 'npm run...' facility, as it just needs a file copy. There is a message at the end of the npm install that @hhellyer added saying:
```
llnode plugin installed, load in lldb with:
(lldb) plugin load .../node_modules/llnode/llnode.so
```
so we could perhaps add a mention of the system plugin directory to that

rnchamberlain:
Message written to console at the end of the npm install is now: 
```
llnode plugin installed, load in lldb with:
(lldb) plugin load .../node_modules/llnode/llnode.so
or copy plugin to lldb system plugin directory
```

gibfahn:
@rnchamberlain is the directory the same across all platforms? It'd be good to either put the directory/directories in the message, or include a link to some document which lists them.

rnchamberlain:
There isn't much user documentation out there for LLDB, an opportunity awaits.... I have documented the system plugin directories in the llnode README as part of this PR. So I could add a github or npmjs.org link to the message. The directories are quite different on Mac vs Linux.

rnchamberlain:
Message written to console at the end of the npm install is now:
```
llnode plugin installed, load in lldb with:
(lldb) plugin load /home/rchamberlain/test/node_modules/llnode/llnode.so
or copy plugin to lldb system plugin directory, see www.npmjs.org/llnode
```
Which is nice, as it provides a clickable link to the help docs etc when the install is done

rnchamberlain:
@indutny @hhellyer when you have a minute could you review this one - thanks!

hhellyer:
LGTM

indutny:
Landed in efab344, thank you!

Issue #73, opened 2017-04-06
something wrong with exec `./gyp_llnode -Dlldb_dir=/usr/lib/llvm-3.9/` On Ubuntu


earlymeme:
```
$ ./gyp_llnode -Dlldb_dir=/usr/lib/llvm-3.9/
Traceback (most recent call last):
  File "./gyp_llnode", line 85, in <module>
    (major, minor), is_clang = compiler_version()
ValueError: need more than 1 value to unpack

```
How to solve it ? My code has been updated.

hhellyer:
@earlymeme - That's checking the output of `cc -version`? What does that return for you? (Which C compiler are you setup to use?)

earlymeme:
@hhellyer - I run `cc --version`, and then tell me that ```cc (Ubuntu 7-20170316-1ubuntu3) 7.0.1 20170316 (experimental) [trunk revision 246203]
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.```
I've found the experimental feature. Is that a problem?

hhellyer:
The code for compiler_version() is:
```
def compiler_version():
  proc = subprocess.Popen(CC.split() + ['--version'], stdout=subprocess.PIPE)
  is_clang = 'clang' in proc.communicate()[0].split('\n')[0]
  proc = subprocess.Popen(CC.split() + ['-dumpversion'], stdout=subprocess.PIPE)
  version = proc.communicate()[0].split('.')
  version = map(int, version[:2])
  version = tuple(version)
  return (version, is_clang)
```
so I guess the next question is what does `cc -dumpversion` return?

earlymeme:
hi @hhellyer - I think this is a bug...When `cc -dumpversion` is only single.And then, a error would throw from `  (major, minor), is_clang = compiler_version()`

hhellyer:
Ok, it's just expecting a two digit version number and getting one digit. I've pushed a fix on a branch.
Could you try installing via:
`npm install hhellyer/llnode#gcc_version_fix`
If that fixes it I'll raise a pull request.

earlymeme:
@hhellyer  It works. Thanks very much. And I see your changed code
```
  # Allow for (e.g.) gcc version 7 (not 7.0)
  version.append("0")
```

your idea is as I thought. And I had changed like yours locally. haha. :)

[...]

PR #78, opened 2017-04-19
Fix issue with gcc 7 confusing gyp_llnode

hhellyer:
cc -dumpversion only reports 7 (not 7.0) on gcc 7, this confuses
the code that expects a two digit version.
This patch just makes sure the version number is padded to at least
two digits so 7 will report 7.0.

### DIFF
--- a/gyp_llnode
+++ b/gyp_llnode
@@ -37,6 +37,8 @@ def compiler_version():
   is_clang = 'clang' in proc.communicate()[0].split('\n')[0]
   proc = subprocess.Popen(CC.split() + ['-dumpversion'], stdout=subprocess.PIPE)
   version = proc.communicate()[0].split('.')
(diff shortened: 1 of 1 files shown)


gyp_llnode
```diff
@@ -37,6 +37,8 @@ def compiler_version():
   is_clang = 'clang' in proc.communicate()[0].split('\n')[0]
   proc = subprocess.Popen(CC.split() + ['-dumpversion'], stdout=subprocess.PIPE)
   version = proc.communicate()[0].split('.')
+  # Allow for (e.g.) gcc version 7 (not 7.0)
+  version.append("0")
```
bnoordhuis: Style: use single quotes for consistency with the surrounding code, and punctuate the comment.

hhellyer:
Quotes fixed.

hhellyer:
This fixes #73

Issue #75, opened 2017-04-12
Update headers for Mac OS X installation to pick up 3.9+ headers


hhellyer:
Xcode recently updated lldb on my system. Now lldb reports:
$ lldb -v
lldb-370.0.40
  Swift-3.1

However this doesn't match up to the actual lldb levels so I forced the installation script to use 3.9 headers. This built and as a result I was able to run v8 findjsobjects *without* needing to run one of the scripts to generate memory regions.

This simplifies the usage of llnode a lot so I will update the npm install scripts. Ideally I need to find a table that maps Xcode versions to 'real' lldb releases from the lldb git repository.

indutny:
Hooray!

indutny:
See: https://github.com/Homebrew/homebrew-core/pull/12608 . Will close this once it lands.

Thanks for update!

hhellyer:
I'll push a fix in the npm build script as well.

For reference I asked about keeping track of the lldb levels in Xcode - the response was that newer levels of lldb should be compatible with older headers. We only need to update the headers if there is a feature we need. (Which there is in this case.)

http://lists.llvm.org/pipermail/lldb-dev/2017-April/012208.html

hhellyer:
I've pushed an update to the npm script. Once this is landed I'll need to update the npm again so that Mac users installing via npm get the benefit.

hhellyer:
v1.4.3 has been published:
https://www.npmjs.com/package/llnode
I can build the plugin with `npm install llnode` and run `findjsobjects` without creating a memory ranges file on MacOS X. Closing this now.

PR #79, opened 2017-04-19
configure.js: Update lldb header downloads.

hhellyer:
The default version of lldb on MacOS now contains the
SBMemoryRegionInfo API allowing us to scan the heap without
using the scripts to generate a segments file.

This simplifies the use of llnode as the v8 find* commands
are now automatically available to the user.

### DIFF
--- a/scripts/configure.js
+++ b/scripts/configure.js
@@ -5,7 +5,7 @@ const fs = require('fs');
 const child_process = require('child_process');
 
 const lldbReleases = {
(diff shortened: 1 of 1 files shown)


hhellyer:
This is for issue #75

hhellyer:
@indutny - This is merged. I'll put an update to the npm up as well so Mac users actually get the benefit. I'll bump the version to 1.4.3

PR #84, opened 2017-05-15
src: support V8 5.8 stack frame markers

bnoordhuis:
On 64 bits systems, V8 stores SMIs (small ints) in the top 32 bits
of a 64 bits word.  Frame markers used to obey this convention but
as of V8 5.8, they are stored as 32 bits SMIs with the top half set
to zero.  Shift the raw value up to make it a normal SMI again.

Fixes test/frame-test.js with the latest node.js v8.0.0-pre, it was
no longer able to detect special JS stack frames like entry and exit
frames, argument adapter frames, etc.

### DIFF
--- a/src/llv8.cc
+++ b/src/llv8.cc
@@ -201,12 +201,26 @@ uint32_t JSFrame::GetSourceForDisplay(bool reset_line, uint32_t line_start,
   return line_start + lines_found;
 }
 
+
+// On 64 bits systems, V8 stores SMIs (small ints) in the top 32 bits of
+// a 64 bits word.  Frame markers used to obey this convention but as of
+// V8 5.8, they are stored as 32 bits SMIs with the top half set to zero.
+// Shift the raw value up to make it a normal SMI again.
+Smi JSFrame::FromFrameMarker(Value value) const {
+  if (v8()->smi()->kShiftSize == 31 && Smi(value).Check() &&
+      value.raw() < 1LL << 31) {
+    value = Value(v8(), value.raw() << 31);
+  }
+  return Smi(value);
+}
+
+
 std::string JSFrame::Inspect(bool with_args, Error& err) {
   Value context =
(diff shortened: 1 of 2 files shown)


indutny:
@bnoordhuis I wonder if we should ask for appropriate post-mortem constants from v8 team?

bnoordhuis:
I don't think there is much the postmortem script can do without major surgery.  The constants are still derived from `StackFrame::Type`, the change is that V8 5.8 stores them differently at runtime.

The comment for `StackFrame::TypeToMarker()` explains it:
```c++
  // Convert a stack frame type to a marker that can be stored on the stack.
  //
  // The marker is an opaque value, not intended to be interpreted in any way
  // except being checked by IsTypeMarker or converted by MarkerToType.
  // It has the same tagging as Smis, so any marker value that does not pass
  // IsTypeMarker can instead be interpreted as a tagged pointer.
  //
  // Note that the marker is not a Smi: Smis on 64-bit architectures are stored
  // in the top 32 bits of a 64-bit value, which in turn makes them expensive
  // (in terms of code/instruction size) to push as immediates onto the stack.
  static int32_t TypeToMarker(Type type) {
    DCHECK_GE(type, 0);
    return (type << kSmiTagSize) | kSmiTag;
  }
```

bnoordhuis:
@indutny Agree/disagree?  Any other comments besides "awesome! fist bump!"?

src/llv8.cc
```diff
@@ -201,12 +201,24 @@ uint32_t JSFrame::GetSourceForDisplay(bool reset_line, uint32_t line_start,
   return line_start + lines_found;
 }
 
+// On 64 bits systems, V8 stores SMIs (small ints) in the top 32 bits of
```
indutny: Nit: please use double newline between functions. It is not enforced by linter, but I prefer this style.

bnoordhuis:
Cheers Fedor, landed with added blank lines in d65792f.

Issue #74, opened 2017-04-11
Unable to read all frames even V8 ones in core dump taken


maziey93:
Hi I have been interested in debugging nodejs in prod for a while but having to install `mdb` was quite an issue, it’s really awesome that there is llnode.

I installed llnode on my mac with brew for practice taking core dumps, analyzing points of failure etc it all worked, so I tried setting up with “a real project” using the simple code [here](https://developer.ibm.com/node/2016/08/15/exploring-node-js-core-dumps-using-the-llnode-plugin-for-lldb/) so everything is predictable but with a docker container. The idea was to get the core dump move it to my machine and debugging, like I would hope to do in production

This is what I got after moving it to my machine, I’d like to know what I might have done wrong

<img width="706" alt="screen shot 2017-04-11 at 2 35 14 pm" src="https://cloud.githubusercontent.com/assets/7893459/24915419/fc7670dc-1ece-11e7-869a-bfce6a271d90.png">

hhellyer:
@maziey93 - Can you post the command line you started lldb with and let me know whether the docker image was Mac or Linux?
I suspect that in the first instance you need to make sure lldb can find the node executable that generated the core dump, otherwise we can't find the symbols we need to understand the v8 object layout.

maziey93:
@hhellyer I started lldb using `lldb node -c <path-to-core>`, the docker image is the official node image for `5.12.0` I know it's some sort of variant of Linux but I'm not sure which it is.

hhellyer:
Ah, so lldb was running in a Mac terminal? The node executable on your path would have been a Mac one rather than a Linux one?

maziey93:
@hhellyer yes I ran lldb in a Mac terminal, but got the core dump from the node docker image which is more or less Linux

hhellyer:
Ok, we could probably use some better examples for this case. But here's what happens for me, using a linux VM that mounts my Mac file system so I can access files in either environment:

To see the same as you:
`> lldb -c ~/work/core.linux.core.on.mac.dmp`
lldb can load this but not find symbols, so even lldb commands like "bt" give very little information.

To make it work, point lldb at the *exact* linux executable you were running when the dump was created:
`> lldb ~/work/node-sdks/node-v6.5.0-linux-x64/bin/node -c ~/work/core.linux.core.on.mac.dmp`
(You can zip it up and move it with the dump if you want.)
[...]

PR #86, opened 2017-05-22
README.md: Improve core dump loading instructions

hhellyer:
Make it clearer how a core dump should be loaded so llnode can obtain all the debug information it needs. I've added the docs to the linux section as (I think) the MacOS core dump actually the executable binaries.

This is for issue #74

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ sudo make install-linux
 
 The llnode plugin can be loaded into LLDB using the `plugin load` command.
 Alternatively it can be installed in the LLDB system plugin directory, in
-which case LLDB will load the plugin automatically on start-up. 
+which case LLDB will load the plugin automatically on start-up.
 
 ### OS X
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -106,6 +106,21 @@ To install the plugin in the LLDB system plugin directory, use the
 npm copy `node_modules/llnode/llnode.so` to
 `/usr/lib/lldb/plugins`.
 
+To use llnode with a core dump the core dump needs to be loaded into lldb along with the exact executable that created the core dump. The executable contains information that lldb and llnode need to make sense of the data in the core dump.
+
+To load the core dump when starting lldb use:
+```
+> lldb ~/work/node-sdks/node-v6.5.0-linux-x64/bin/node -c ~/work/core
```
bnoordhuis: Maybe use fictitious paths here (e.g. `/path/to/bin/node` and `/path/to/core`) or clarify that the paths are example paths.

As well, I'd replace the `>` with a `$` to make it clear it's meant to be run from the shell.

Lastly, can you wrap lines at 80 columns?  Cheers.

hhellyer: I've updated the paths and wrapped the lines.
Looking at the rest of the file $ isn't used for a prompt else where commands are left bare with no prompt character so I've gone with that convention.

I took a look at a few other projects and couldn't see a clear preference. One advantage here is that the longer multiline installation commands in this README.md use inline comments with # and those whole sections can just be pasted into a shell to do the complete installation.

bnoordhuis:
@hhellyer You want to land this?

hhellyer:
@bnoordhius - Yep, I need to do an update to the npm (PR #94 should be published!) so I'll land the various readme updates and do that.

Issue #77, opened 2017-04-12
npm install -g llnode fails


rnchamberlain:
It's common practice to install tools using the` -g `option, but this seems to fail with llnode:

```
rchamberlain@ThinkCentre-M57p:~/test$ npm install -g llnode

> llnode@1.4.1 preinstall /home/rchamberlain/node-v7.7.3-linux-x64/lib/node_modules/.staging/llnode-4a7a78c7
> node scripts/configure.js

Build dir is: /home/rchamberlain/node-v7.7.3-linux-x64/lib/node_modules/.staging/llnode-4a7a78c7
Checking for headers, version is 3.9
Installing llnode for /usr/lib/llvm-3.9/bin/lldb, lldb version 3.9
Linking lldb to include directory /usr/lib/llvm-3.9
npm ERR! Linux 4.4.0-72-generic
npm ERR! argv "/home/rchamberlain/node-v7.7.3-linux-x64/bin/node" "/home/rchamberlain/node-v7.7.3-linux-x64/bin/npm" "explore" "node-gyp" "pwd"
npm ERR! node v7.7.3
npm ERR! npm  v4.1.2

npm ERR! It doesn't look like node-gyp is installed.
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     /home/rchamberlain/node-v7.7.3-linux-x64/lib/node_modules/npm/npm-debug.log
npm ERR! weird error 1
child_process.js:512
    throw err;
    ^
Error: Command failed: npm -g explore npm npm explore node-gyp pwd
```

Problem seems to be this line in the config.js script: https://github.com/nodejs/llnode/blob/master/scripts/configure.js#L89

Using `npm install llnode `without the `-g` option works fine.

hhellyer:
PR #94 has been landed and published to https://www.npmjs.com/package/llnode - Closing this now.

PR #85, opened 2017-05-18
readme: fix typo

lpinca:
This fixes a typo in README.md.

PR #94, opened 2017-06-07
Fix for installing llnode as a global

No9:
When `npm explore` is ran with a `-g` option the default folder is set to `npm.globalDir` 
which is the root `$GLOBAL/lib/node_modules` folder containing npm. 
But when it's ran in local mode it uses path.resolve(npm.prefix, 'node_modules') 
See https://github.com/npm/npm/blob/latest/lib/npm.js#L395 
This is causing the resolution of node-gyp to fail in global mode. 

This patch tests the npm environment variable `npm_config_global` to see if it's a global install and appends the required sub directories to get it to work.

Tested on Ubuntu 16.04 with 
```
$ git clone https://github.com/no9/llnode
$ cd llnode
$ npm install -g 
```

Fixes: https://github.com/nodejs/llnode/issues/77

### DIFF
--- a/scripts/configure.js
+++ b/scripts/configure.js
@@ -81,13 +81,20 @@ if (lldbHeadersBranch != undefined) {
 console.log(`Linking lldb to include directory ${lldbIncludeDir}`);
 fs.symlinkSync(lldbIncludeDir, 'lldb');
 
+// npm explore has a different root folder when using -g 
+// So we are tacking on the extra the additional subfolders
+var gypSubDir = 'node-gyp';
+if(process.env.npm_config_global) {
+    gypSubDir = 'npm/node_modules/node-gyp';
+}
+
 // Initialize GYP
 // We can use the node-gyp that comes with npm.
 // We can locate it with npm -g explore npm npm explore node-gyp pwd
 // It might have been neater to make node-gyp one of our dependencies
 // *but* they don't get installed until after the install step has run.
 var gypDir = child_process.execFileSync('npm',
-  ['-g', 'explore', 'npm', 'npm', 'explore', 'node-gyp', 'pwd'],
+  ['-g', 'explore', 'npm', 'npm', 'explore', gypSubDir, 'pwd'],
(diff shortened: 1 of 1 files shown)


hhellyer:
The fix looks good to me, I've tested locally and both local and global installs now work.

rnchamberlain:
LGTM - tested on Ubuntu 16.04. It's very useful to have` install -g` working for this - thanks!

scripts/configure.js
```diff
@@ -81,13 +81,20 @@ if (lldbHeadersBranch != undefined) {
 console.log(`Linking lldb to include directory ${lldbIncludeDir}`);
 fs.symlinkSync(lldbIncludeDir, 'lldb');
 
+// npm explore has a different root folder when using -g 
+// So we are tacking on the extra the additional subfolders
+var gypSubDir = 'node-gyp';
+if(process.env.npm_config_global) {
```
indutny: Nit: style, space between `if` and `(`

scripts/configure.js
```diff
@@ -81,13 +81,20 @@ if (lldbHeadersBranch != undefined) {
...
+    gypSubDir = 'npm/node_modules/node-gyp';
```
indutny: Nit: style, 2-space indent.

No9:
Thanks for the reviews - nits are resolved and PR updated.

indutny:
Landed in fb24599, thank you! Sorry for delay.

hhellyer:
@indutny - Since this is a fix to npm installation I'll update the version to 1.5.1 and publish a new version of the npm on Monday, unless you object!
(The brew installation probably doesn't need an update for this.)

rnchamberlain:
@indutny @hhellyer could we land the README fixes as well, then publish? https://github.com/nodejs/llnode/pull/85 and https://github.com/nodejs/llnode/pull/86

hhellyer:
@rnchamberlain - Yep, I'll do that today.

Issue #99, opened 2017-06-19
v8 bt skips some frames


brendangregg:
Maybe there's a good reason for this, but notice we're skipping frames 1-4:

```
(lldb) v8 bt
 * thread #1: tid = 13, 0x00007fda23874d44 libpthread.so.0`__GI___pthread_mutex_lock + 4, name = 'node', stop reason = signal SIGSEGV
  * frame #0: 0x00007fda23874d44 libpthread.so.0`__GI___pthread_mutex_lock + 4
    frame #5: 0x00007fda20ffe212 node`v8::internal::FunctionCallbackArguments::Call(void (*)(v8::FunctionCallbackInfo<v8::Value> const&)) + 290
[...]
```

They do show up with the normal bt:

```
(lldb) bt
* thread #1, name = 'node', stop reason = signal SIGSEGV
  * frame #0: 0x00007fda23874d44 libpthread.so.0`__GI___pthread_mutex_lock + 4
    frame #1: 0x00007fda20b0ed19 node
    frame #2: 0x00007fda20ac5e39 node
    frame #3: 0x00007fda20a9dd4b node
    frame #4: 0x00007fda20a9c9f6 node
    frame #5: 0x00007fda20ffe212 node`v8::internal::FunctionCallbackArguments::Call(void (*)(v8::FunctionCallbackInfo<v8::Value> const&)) + 290
[...]
```

Should "v8 bt" be skipping them?

bnoordhuis:
I've been looking at this.  llnode skips frames it can't link to either a C++ symbol or a JS stack frame.

It's easy to change but then stack frames for V8 builtins also show up (because llnode does not know how to identify them) and there are often a great number of them.

I added a heuristic to skip frames whose PC is inside a WX memory segment on the assumption that it's almost certainly jitted code.  The only problem is that it depends on `SBProcess::GetMemoryRegions()` so I don't think it's going to work with older lldb versions.

Before I move on with this, does anyone have a better idea?

rnchamberlain:
We had a brief discussion on skipped frames last year, and there was interest in seeing all frames, even if the information available was minimal, eg V8 transition frames.

Re dependency on `GetMemoryRegions()`, as that API is in lldb 3.9, which is the stable release now, I don't think that should be a worry.

bnoordhuis:
Sounds good.  I've opened #104, it prints them as `<builtin>`.

PR #104, opened 2017-06-21
src: print builtins and unnamed stack frames

bnoordhuis:
Previously, `v8 bt` would exclude frames that didn't map to a C++ symbol
or a JS stack frame.  llnode does not currently know how to identify the
stack frames of V8 builtins so those were omitted as well.

This commit makes those stack frames visible and introduces a heuristic
(in lldb >= 3.9) where frames whose PC is inside a WX memory segment are
assumed to belong to V8 builtins.

Fixes: #99

First commit is cleanup, second commit fixes a test, third one is the fix for #99.

### DIFF
--- a/src/llnode.cc
+++ b/src/llnode.cc
@@ -17,6 +17,7 @@ using lldb::SBDebugger;
 using lldb::SBError;
 using lldb::SBExpressionOptions;
 using lldb::SBFrame;
+using lldb::SBMemoryRegionInfo;
 using lldb::SBStream;
 using lldb::SBSymbol;
 using lldb::SBTarget;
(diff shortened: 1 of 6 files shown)


src/llnode.cc
```diff
@@ -97,30 +110,37 @@ bool BacktraceCmd::DoExecute(SBDebugger d, char** cmd,
...
 
-    // V8 frame
-    v8::Error err;
-    v8::JSFrame v8_frame(&llv8, static_cast<int64_t>(frame.GetFP()));
-    std::string res = v8_frame.Inspect(true, err);
-
-    // Skip invalid frames
-    if (err.Fail()) continue;
+#ifdef LLDB_SBMemoryRegionInfoList_h_
+    // Heuristic: a PC in WX memory is almost certainly a V8 builtin.
```
indutny: Should we ask V8 team for some post-mortem information to be able to identify them?

bnoordhuis: I don't know how yet.  Builtins are generated dynamically, they don't have fixed addresses or frame markers or anything.

I have a proof-of-concept that looks them up through the isolate pointer but it still feels a little brittle.  Still pondering.

indutny: Makes sense then.

indutny:
Thanky ou

brendangregg:
+1 Tested, works, thanks.

```
(lldb) v8 bt
 * thread #1: tid = 13, 0x00007fda23874d44 libpthread.so.0`__GI___pthread_mutex_lock + 4, name = 'node', stop reason = signal SIGSEGV
  * frame #0: 0x00007fda23874d44 libpthread.so.0`__GI___pthread_mutex_lock + 4
    frame #1: 0x00007fda20b0ed19 node
    frame #2: 0x00007fda20ac5e39 node
    frame #3: 0x00007fda20a9dd4b node
    frame #4: 0x00007fda20a9c9f6 node
    frame #5: 0x00007fda20ffe212 node`v8::internal::FunctionCallbackArguments::Call(void (*)(v8::FunctionCallbackInfo<v8::Value> const&)) + 290
[...]
```

That is an improvement - better than skipping!

src/llnode.cc
```diff
@@ -11,7 +11,20 @@
...
 namespace llnode {
 
-using namespace lldb;
+using lldb::SBCommandInterpreter;
+using lldb::SBCommandReturnObject;
+using lldb::SBDebugger;
+using lldb::SBError;
+using lldb::SBExpressionOptions;
+using lldb::SBFrame;
+using lldb::SBMemoryRegionInfo;
```
hhellyer: This line breaks on lldb 3.8, which is the default lldb on Ubuntu. (Works fine on Mac and later levels of lldb.)
We use `#ifdef LLDB_SBMemoryRegionInfoList_h_` elsewhere.

bnoordhuis: Right you are.  Updated.

bnoordhuis:
@hhellyer Next time, can you not squash commits unless they're marked as such?  It buries the actual change in unrelated changes.

As well, can you use full URLs in Fixes and PR-URL tags?  Thanks.

Issue #89, opened 2017-05-31
Reading contents of an Array/ArrayBufferView


joyeecheung:
Hi, at the moment when doing `v8 inspect <address-of-an-array>` llnode would only print the length (and offset & backstore address) of an Array (ArrayBufferView). Are there any support for reading the elements of an Array/ArrayBufferView? Thanks.

bnoordhuis:
Do you mean Array or ArrayBuffer?  If the latter: not at the moment.  You can take the address of the backing store and print that like you would with normal memory.

We could pretty-print ArrayBuffers but we'd need to do something sensible for large arrays, like print only the first 100 elements.

If you want to work on it, the logic is in `JSArrayBuffer::Inspect()` and `JSArrayBufferView::Inspect()` in src/llv8.cc.  Questions?  I'll try to answer them.

joyeecheung:
@bnoordhuis Thanks for the answer, I actually meant both, but I just realized it does support arrays, I guess the arrays that I've tried somehow errorred in one of the steps of `JSObject::InspectElements` so the elements didn't show up.

I think we could do a new command for array-like objects which also takes indexes/ranges as operand? I'd be happy to work on it.

bnoordhuis:
A flag sounds good but the default should probably be conservative for large arraybuffers because those can easily be 100s of MBs and you don't usually want to see every last byte printed.

As well, it would be nice if typed arrays printed their values as their natural types (uint8, int32, float64, etc.) instead of simply dumping out the `.buffer` property but that's a bigger undertaking.

bnoordhuis:
@joyeecheung Just curious, were you able to make any progress on this?

joyeecheung:
@bnoordhuis I have implemented `--array-length` for normal arrays but have not found the time to do the same for array buffers. I will open a PR for what have been done, can probably implement the other parts this weekend.

joyeecheung:
@hhellyer The implementation for `ArrayBuffer` and `ArrayBufferView` has not been completed yet, reopening..

hhellyer:
@joyeecheung - Ah, I suspect that's because the commit comment referenced this with "Fix", sorry about that!

PR #105, opened 2017-06-21
src: implement `v8 inspect --array-length` for large arrays

joyeecheung:
Also added tests for `--string-length` since they are pretty similar...

Refs: https://github.com/nodejs/llnode/issues/89

cc:/ @bnoordhuis

### DIFF
--- a/src/llv8.cc
+++ b/src/llv8.cc
@@ -1,5 +1,6 @@
 #include <assert.h>
 
+#include <algorithm>
 #include <cinttypes>
 
 #include "llv8-inl.h"
@@ -1283,9 +1284,18 @@ std::string JSObject::InspectElements(Error& err) {
   Smi length_smi = elements.Length(err);
   if (err.Fail()) return std::string();
 
+  int64_t length = length_smi.GetValue();
+  return InspectElements(length, err);
+}
+
+
+std::string JSObject::InspectElements(int64_t length, Error& err) {
+  HeapObject elements_obj = Elements(err);
+  if (err.Fail()) return std::string();
+  FixedArray elements(elements_obj);
+
   InspectOptions options;
 
-  int64_t length = length_smi.GetValue();
   std::string res;
   for (int64_t i = 0; i < length; i++) {
     Value value = elements.Get<Value>(i, err);
@@ -1843,7 +1853,8 @@ std::string JSArray::Inspect(InspectOptions* options, Error& err) {
 
   std::string res = "<Array: length=" + std::to_string(length);
   if (options->detailed) {
-    std::string elems = InspectElements(err);
+    int64_t display_length = std::min<int64_t>(length, options->array_length);
+    std::string elems = InspectElements(display_length, err);
     if (err.Fail()) return std::string();
 
     if (!elems.empty()) res += " {\n" + elems + "}";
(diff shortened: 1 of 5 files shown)


src/llv8.cc
```diff
@@ -1838,10 +1838,10 @@ v8::Value JSObject::GetArrayElement(int64_t pos, Error& err) {
 }
 
 std::string JSArray::Inspect(InspectOptions* options, Error& err) {
-  Smi length = Length(err);
+  int64_t length = GetArrayLength(err);
```
joyeecheung: The original code would call `NameDictionary::Length()` instead so the length displayed would count an additional entry (`length`).

src/llv8.cc
```diff
@@ -1838,12 +1847,16 @@ v8::Value JSObject::GetArrayElement(int64_t pos, Error& err) {
...
   if (err.Fail()) return std::string();
 
-  std::string res = "<Array: length=" + length.ToString(err);
+  std::string res = "<Array: length=" + std::to_string(length);
   if (options->detailed) {
-    std::string elems = InspectElements(err);
+    int64_t display_length = length;
+    if (length > options->array_length) {
+      display_length = options->array_length;
+    }
```
bnoordhuis: `int64_t display_length = std::min(length, options->array_length);`?  Arguably a little easier to read.

If not that, can you `s/length/display_length/` in the comparison?

test/inspect-test.js
```diff
@@ -72,6 +74,16 @@ tape('v8 inspect', (t) => {
     t.ok(/.other-key=[^\n]*<String: "ohai">/.test(lines),
          '.other-key property');
 
+    const arrayMatch = lines.match(
+    /.array=(0x[0-9a-f]+):<Array: length=6>/);
```
bnoordhuis: Can you indent by 4 spaces here and elsewhere?  (Better, indent but break after the `=`.)

joyeecheung:
@bnoordhuis @indutny  Thanks for the review, updated per https://github.com/nodejs/llnode/pull/105#pullrequestreview-45466113 PTAL.

I have tested this with v4.8.3, v6.10.0 and v8.0.0 on MacOS 10.12.4, all tests have passed.

test/inspect-test.js
```diff
@@ -72,6 +74,16 @@ tape('v8 inspect', (t) => {
     t.ok(/.other-key=[^\n]*<String: "ohai">/.test(lines),
          '.other-key property');
 
+    const arrayMatch = 
+      lines.match(/.array=(0x[0-9a-f]+):<Array: length=6>/);
```
bnoordhuis: Can you use 4 space indent here and below?

joyeecheung:
@bnoordhuis Fixed the indentation, PTAL.

test/inspect-test.js
```diff
@@ -96,6 +108,40 @@ tape('v8 inspect', (t) => {
...
+  });
+
+  sess.linesUntil(/}>/, (lines) => {
+    lines = lines.join('\n');
+    t.notEqual(
+        lines.indexOf('<Array: length=6'),
+        -1,
+        'array length');
+    t.ok(
+        lines.match(/\[5\]=0x[0-9a-f]+:<function: Class at .+\.js:\d+:\d+>}>$/),
```
cjihrig: It doesn't look like you're using the result of `match()` here and on line 143. It might be better to use `test()` instead.

PR #106, opened 2017-06-23
src: don't crash with no arguments to print

cjihrig:
This commit prevents a segfault on `v8 print`.

I'll be adding a test as well, but it requires some changes to `test/common.js` first.

### DIFF
--- a/src/llnode.cc
+++ b/src/llnode.cc
@@ -130,7 +130,7 @@ bool BacktraceCmd::DoExecute(SBDebugger d, char** cmd,
 
 bool PrintCmd::DoExecute(SBDebugger d, char** cmd,
                          SBCommandReturnObject& result) {
-  if (*cmd == nullptr) {
+  if (cmd == nullptr || *cmd == nullptr) {
     if (detailed_) {
       result.SetError("USAGE: v8 inspect [flags] expr\n");
     } else {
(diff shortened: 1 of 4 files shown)


src/llnode.cc
```diff
@@ -130,7 +130,7 @@ bool BacktraceCmd::DoExecute(SBDebugger d, char** cmd,
 
 bool PrintCmd::DoExecute(SBDebugger d, char** cmd,
                          SBCommandReturnObject& result) {
-  if (*cmd == NULL) {
+  if (cmd == NULL || *cmd == NULL) {
```
bnoordhuis: Can you use nullptr here?  Seems you stumbled on one of the three places that still uses NULL (vs. ~50 instances of nullptr.)

Are other commands affected too?  A quick grep suggests some, like `list` and `findrefs`, don't check `cmd == nullptr` either.

cjihrig:
Replaced the remaining uses of `NULL` and updated the other commands in a similar fashion. I've got four test cases to add, but have to update the test `Session` to expose `stderr` first.

cjihrig:
OK, I allowed stderr to be scripted the same way that stdout currently is. Then, I added a test for the usage messages. Those commands segfault on master, and work correctly with the fixes in this PR.

cjihrig:
@bnoordhuis what is the policy for PRs getting landed in this repo? I technically have commit rights and can land this myself, but I don't want to step on any toes either.

bnoordhuis:
Go ahead and land it, no toes to worry about.

hhellyer:
@bnoordhuis @cjihrig - I can land things but I'm not sure whether I should be able to in a Node.js repo or whether my access ended up being grandfathered in when @indutny moved the repo from his space to the nodejs org. So I don't usually unless the PR has had a couple of LGTMs and been waiting a little while!
Things are building up though - I'm happy to clean up the backlog if you guys want.

bnoordhuis:
@hhellyer Go for it! :-)

PR #108, opened 2017-06-29
Add an llnode command script that invokes lldb with the plugin loaded.

hhellyer:
Since @No9 fixed global installation it's more practical to add an llnode command using a `bin` entry in package.json. This is a work in progress PR to get some feedback on whether anyone else thinks it's a good idea.

This implementation is fairly simple, it just wrappers lldb and pre-loads the plugin. (It also changes the prompt so you know it's worked.) Any other argument is just passed straight through so all the existing lldb options can be passed.

This allows the debug workflows:
`$ npm install -g llnode`
followed by:
`$ llnode node -c core` for postmortem debugging.
or
`$ llnode -- node --abort_on_uncaught_exception test.js` for JavaScript debug.
or
`$ llnode -- node test_for_crashing_native_npm.js` for crashes when developing npm modules containing native code.
(There are probably quite a few other use cases. Suggestions are welcome so they can be documented.)

Hopefully this will improve the ease of use and make it easier to recommend llnode to a wider range of Node.js developers. I know the brew installer achieves a similar thing but it's not available for Linux and just installing an npm package feels more "Node-y" which might make a wider audience more likely to use llnode.

Missing from this PR is a re-write of the documentation but I wanted to get some feedback from everyone else before I sat down to do that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,23 @@
 
 [![Build Status](https://secure.travis-ci.org/nodejs/llnode.png)](http://travis-ci.org/nodejs/llnode)
 
-Node.js v4.x-v6.x C++ plugin for [LLDB](http://lldb.llvm.org) - a next generation, high-performance debugger.
+Node.js v4.x-v8.x C++ plugin for the [LLDB](http://lldb.llvm.org) debugger.
+
+The llnode plugin adds the ability to inspect JavaScript stack frames, objects,
+source code and more to the standard C/C++ debugging facilities when working
+with Node.js processes or core dumps in LLDB.
+
+### Quick start
+
+Start an lldb session with the llnode plugin pre-loaded:
+
+```bash
+npm install -g llnode
+llnode node -c core
+```
+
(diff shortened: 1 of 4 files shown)


hhellyer:
I've updated the README.md to (hopefully) explain the new command and make the instructions more use to someone who does `npm install -g llnode` and just wants to get debugging. (It's a bit unfortunate you can't seem to specify a different readme for npmjs.org and github.com.)

If you spot problems I'm happy for reviewers to edit the README.md on this branch, it should allow that and it's probably quicker than round tripping fixes to typos. (We can squash on merge.)

hhellyer:
I had to push a change as the generated shell script started with a leading blank line which caused `#!/bin/sh` to be ignored.

I think this is ready to go now, once I land it does anyone object to me bumping the version to 1.6 and updating the npm on npmjs.org? We've landed quite a lot of PR's since the last version so it would be worth doing anyway and this one doesn't make much sense without an npm update to finish it off.

PR #110, opened 2017-06-30
Allow npm install to work in git clone.

hhellyer:
It looks like the travis CI build does an npm install in a git checkout.
I've tested that locally and it fails due to some of the directories already existing. This PR tries to fix that and also means that running npm install in a local git clone should work.

I think travis will build PR's automatically so I may add changes if this does not resolve the issue.


hhellyer:
The CI build isn't quite clean - it builds but `npm test` fails. I have a fix for that but I'll get it reviewed on a different branch and in the meantime I'll merge these since they are approved.

(I got started on this as PR #107 fixes the build badge but the build is still broken.)

PR #111, opened 2017-07-03
WIP - test: Fix the path to llnode plugin

hhellyer:
Load the llnode plugin from the location npm install creates it at.

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -19,9 +19,9 @@ if (process.platform === 'darwin')
 else if (process.platform === 'windows')
   pluginName = 'llnode.dll';
 else
-  pluginName = path.join('lib.target', 'llnode.so');
+  pluginName = 'llnode.so';
 
-exports.llnodePath = path.join(exports.buildDir, pluginName);
+exports.llnodePath = pluginName;
 
 function SessionOutput(session, stream) {
   EventEmitter.call(this);
--- a/test/common.js
+++ b/test/common.js
@@ -14,14 +14,36 @@ exports.core = path.join(os.tmpdir(), 'core');
 exports.ranges = exports.core + '.ranges';
 
 let pluginName;
-if (process.platform === 'darwin')
+if (process.platform === 'darwin') {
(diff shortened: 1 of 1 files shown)


hhellyer:
This fix causes `npm test` to load the plugin library from the location `npm install` creates it at. This is probably correct but it is a different location to the location `./gyp_llnode && make -C out/` will use so now the tests fail after building that way. (Previously they worked that way but failed if `npm install` was used to build.)

This gets the test code running on travis but the tests are still failing:
https://travis-ci.org/nodejs/llnode/jobs/249579889
(I'll see if I can figure out why quickly and maybe add another commit to this.)

cjihrig:
> This is probably correct but it is a different location to the location ./gyp_llnode && make -C out/ will use so now the tests fail after building that way.

Would it make sense to try to detect the location in `common.js`, or maybe even use an environment variable to define the location.

hhellyer:
@cjihrig - It might do, there's not that many places to look. I guess the problem is which to pick if both are found. (That could just result in a test failure that reminds you to delete one of them though!)

cjihrig:
We could emit a warning or complain loudly in some other way if they are both found.

hhellyer:
Slightly awkwardly the travis build runs npm install which builds the plugin then runs the _travis make target which runs gyp_llnode and compiles the plugin. This change blocks all the tests running on travis:
https://travis-ci.org/nodejs/llnode/jobs/249994477

I'll figure out what to do next, I have to push the changes to a branch on a PR so that travis picks them up though so I've added a WIP tag to this PR!

joyeecheung:
Travis is now working with the plugin built with `npm install`. Should this be closed, or updated to fallback to the plugin in `out/Release` when there isn't one in the project directory, or something else?

bnoordhuis:
Ping @hhellyer.  Can you update or close?

hhellyer:
I'll close, it's been a bit too long since I looked at this and I'm not in a position to update it at the moment. Sorry for leaving it lying around!

Issue #97, opened 2017-06-19
findrefs performance improvement


brendangregg:
Currently findrefs has to rescan the entire heap each time. Could an index or some cache be added to improve performance of this when using it multiple times?

bnoordhuis:
Are you experiencing performance issues?  If so, can you provide some background on the size of the core file, number of objects in it, etc.?

brendangregg:
I'm hearing about latency second hand. Maybe the first step here is to verify that this really is an issue and provide such numbers for context, including how slow the command is. Might need #98 fixed first to really see some latency.

bnoordhuis:
I'll close this for now.  Happy to investigate once there is some concrete data.

PR #114, opened 2017-07-10
Improved findrefs execution speed after it runs the first time

mmarchini:
With those changes, llscan will keep a cache of searched references and
will use this cache to avoid rescanning the entire heap each time.
This will improve performance when executing findrefs multiple times.
Fixes: https://github.com/nodejs/llnode/issues/97

### DIFF
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -1146,10 +1146,8 @@ bool LLScan::ScanHeapForObjects(lldb::SBTarget target,
   // LLNODE_RANGESFILE with data for the new dump or things won't match up).
   if (target_ != target) {
     ClearMemoryRanges();
-    mapstoinstances_.clear();
-    references_by_value_.clear();
-    references_by_property_.clear();
-    references_by_string_.clear();
+    ClearMapsToInstances();
+    ClearReferences();
     target_ = target;
   }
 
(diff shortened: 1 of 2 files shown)


mmarchini:
I realized that this solution is very memory consuming, so I'll update my PR tomorrow with a much less consuming version, and also with improvements on the cache initialization.

brendangregg:
I tested the first version, which works, thanks.

Minor comment: I think the whitespace style is inconsistent with the rest of llnode. eg: ```if(references==nullptr) {``` should be ```if (references == nullptr) {```.

indutny:
`make format` should be able to fix style nits.

hhellyer:
Looks good, I'm happy to do a more detailed review once you've pushed your other improvements.

mmarchini:
Just updated the PR with a new solution, very lightweight on memory and still very fast. It also adds almost no overhead to the first execution.

And I profiled of both versions to compare speed using a core with ~180Mb:

Findrefs execution time on master:
findrefs execution time on master

Findrefs execution time with those changes:
findrefs execution time with those changes

What do you guys think?

mmarchini:
Forgot to mention: it will create the entire cache after running `findrefs` the first time **for each type of scan** (by reference, by property and by string).

hhellyer:
@mmarchini - I think you still have a memory leak when the user changes targets. When you clear the maps you remove the pointers they contain not the memory pointed to by those pointers (created by `new ReferencesVector`).
You can look at the function `LLScan::ClearMapsToInstances()` at the bottom of llscan.cc as an example of how to free them, except it turns out that `LLScan::ClearMapsToInstances()` isn't being called and we have a bit of a memory leak there. I'll raise a separate issue for that!
(As an alternative you could also probably store `std::shared_ptr<ReferencesVector>`s in your maps, that might nice in C++ terms although there will be a small footprint overhead to using shared pointers as they have to store a reference count.)

There are also a few compiler warnings introduced by the functions `AreReferencesLoaded()` and `GetReferences()` where the base implementations don't return anything, at least on Mac, could you tidy those up?

Other than that it all looks good to me and should make life easier when running multiple queries against a dump.

mmarchini:
I think `std::shared_ptr` will add unnecessary overhead on both speed and memory usage, so I ended up destroying those references manually. I took the liberty of fixing mapsofinstances_ memory leaks too (if you prefere I'll create another PR for that).

Those compiler warnings are not appearing for me (I'm on Ubuntu 16.04), but I think they're fixed now.

hhellyer:
Merged, sorry for the delay.

PR #120, opened 2017-08-15
src: Add support for big endian systems

hhellyer:
Read values correctly converting for endianess as required.
Generally this just means using LLDB's `ReadUnsignedFromMemory` function but there are a few places where we need optimise reading values and where we only want one byte from a larger field where we need to account for endianess ourselves.

I've tested this with z/Linux core dumps and it enables commands like `v8 findjsobjects`, `findjsinstances` and `v8 inspect` to work. (I've used llnode with these changes to open the z/Linux dumps on my Mac.)

`v8 bt` does not work but that is due to a separate issue with how v8 stacks are implemented on non-x86 systems.

I've verified the tests still pass which confirms these changes don't break little endian support.

### DIFF
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -17,6 +17,7 @@
 
 namespace llnode {
 
+using lldb::ByteOrder;
 using lldb::SBCommandReturnObject;
 using lldb::SBDebugger;
 using lldb::SBError;
@@ -1204,12 +1205,19 @@ bool LLScan::ScanHeapForObjects(lldb::SBTarget target,
   return true;
 }
 
+inline static ByteOrder GetHostByteOrder() {
+  union {
+    uint8_t a[2];
+    uint16_t b;
+  } u = {{0, 1}};
+  return u.b == 1 ? ByteOrder::eByteOrderBig : ByteOrder::eByteOrderLittle;
+}
 
 void LLScan::ScanMemoryRanges(FindJSObjectsVisitor& v) {
   bool done = false;
 
   const uint64_t addr_size = process_.GetAddressByteSize();
-  ByteOrder process_byte_order = process_.GetByteOrder();
+  bool swap_bytes = process_.GetByteOrder() != GetHostByteOrder();
 
   // Pages are usually around 1mb, so this should more than enough
   const uint64_t block_size = 1024 * 1024 * addr_size;
@@ -1261,12 +1269,12 @@ void LLScan::ScanMemoryRanges(FindJSObjectsVisitor& v) {
 
         if (addr_size == 4) {
           value = *reinterpret_cast<uint32_t*>(&block[j]);
-          if (process_byte_order != HOST_BYTE_ORDER) {
+          if (swap_bytes) {
             value = __builtin_bswap32(value);
           }
         } else if (addr_size == 8) {
           value = *reinterpret_cast<uint64_t*>(&block[j]);
-          if (process_byte_order != HOST_BYTE_ORDER) {
+          if (swap_bytes) {
             value = __builtin_bswap64(value);
           }
         } else {
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -1264,8 +1264,7 @@ void LLScan::ScanMemoryRanges(FindJSObjectsVisitor& v) {
           if (process_byte_order != HOST_BYTE_ORDER) {
             value = __builtin_bswap32(value);
           }
-        }
-        else if (addr_size == 8) {
+        } else if (addr_size == 8) {
           value = *reinterpret_cast<uint64_t*>(&block[j]);
           if (process_byte_order != HOST_BYTE_ORDER) {
             value = __builtin_bswap64(value);
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -1209,6 +1209,7 @@ void LLScan::ScanMemoryRanges(FindJSObjectsVisitor& v) {
   bool done = false;
 
   const uint64_t addr_size = process_.GetAddressByteSize();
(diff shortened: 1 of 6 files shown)


src/llv8-inl.h
```diff
@@ -83,6 +83,12 @@ inline int64_t HeapObject::GetType(Error& err) {
 inline int64_t Map::GetType(Error& err) {
   int64_t type = LoadField(v8()->map()->kInstanceAttrsOffset, err);
   if (err.Fail()) return -1;
+
+  if( v8()->process_.GetByteOrder() != HOST_BYTE_ORDER) {
+    // The type is only one byte within the instance attrs.
+    type = type >> 0x30;
```
indutny: Shouldn't this be `0x38`?

src/llv8-inl.h
```diff
@@ -150,16 +156,16 @@ ACCESSOR(Map, InstanceDescriptors, map()->kInstanceDescriptorsOffset,
...
 }
 
 inline int64_t Map::InObjectProperties(Error& err) {
   return LoadField(v8()->map()->kInObjectPropertiesOffset, err) & 0xff;
 }
 
 inline int64_t Map::InstanceSize(Error& err) {
-  return (LoadField(v8()->map()->kInstanceSizeOffset, err) & 0xff) *
-         v8()->common()->kPointerSize;
+  return v8()->LoadUnsigned(LeaField(v8()->map()->kInstanceSizeOffset), 1, err)
```
indutny: Let's run `make format` against this. Putting operator on the next line is definitely not what our style allows at the moment.

src/llscan.cc
```diff
@@ -1242,12 +1243,20 @@ void LLScan::ScanMemoryRanges(FindJSObjectsVisitor& v) {
...
+            value = __builtin_bswap32(value);
+          }
+        }
+        else if (addr_size == 8) {
           value = *reinterpret_cast<uint64_t*>(&block[j]);
-        else
+          if (process_byte_order != HOST_BYTE_ORDER) {
+            value = __builtin_bswap64(value);
+          }
+        } else {
```
bnoordhuis: Since you're here anyway, maybe rewrite this so it won't run afoul of alignment issues on non-x86.  I mean this:
```c++
uint32_t tmp;
memcpy(&tmp, &block[j], sizeof(tmp));
value = tmp;
// do byte swap if needed
```
Fine to do in a separate commit.

src/llv8.h
```diff
@@ -7,6 +7,14 @@
 
 #include "src/llv8-constants.h"
 
+using lldb::ByteOrder;
+
+#if __BYTE_ORDER == __LITTLE_ENDIAN
+#define HOST_BYTE_ORDER ByteOrder::eByteOrderLittle
+#else
+#define HOST_BYTE_ORDER ByteOrder::eByteOrderBig
+#endif
```
bnoordhuis: Depending on `__BYTE_ORDER` is a little fragile, you often end up with a tower of #ifdefs to include the right file.  It's not necessary either because you can derive it at run-time:
```c++
inline bool IsBigEndian() {
  union { uint8_t a[2]; uint16_t b; } u = {{0, 1}};
  return u.b == 1;
}
```

src/llv8-inl.h
hhellyer: Since "because that works" wasn't a good answer I found the v8 comments explaining the way it works:
```
// Order instance type and bit field together such that they can be loaded
// together as a 16-bit word with instance type in the lower 8 bits regardless
// of endianess. Also provide endian-independent offset to that 16-bit word.
```
I've updated the code to read a 16 bit unsigned value and that works both ways without a special case.

src/llscan.cc
hhellyer: I think it's probably worth raising a separate issue for this. I'm a bit concerned doing a memcpy for every word would eliminate some of the performance gains from reading blocks of memory instead of individual words.

src/llv8.h
hhellyer: I'm changing to the little endian define to use `__ORDER_LITTLE_ENDIAN__` since `__BYTE_ORDER__` and `__ORDER_LITTLE_ENDIAN__` are both built in to gcc and clang. You can compile the program below as is (with no extra includes) so these should be available by default.
```
int main(int argc, char **argv)
{
    return __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__;
}
```
I don't think there's any point working it out at run time if you can work it out at compile time. (The compiler might be clever and work it out anyway.

rnchamberlain:
Re-tested with latest commits above. Big-endian core dump from zLinux opens OK using LLDB 4.0 on a Linux/Intel box, and v8 findjsobjects and findjsinstances commands work fine.

src/llscan.cc
bnoordhuis: gcc and clang lower it to `__builtin_memcpy()` which emits simple loads and stores for word-sized copies (on architectures where alignment isn't an issue.)  I'd be surprised if it was in any way slower.

src/llv8.h
```diff
@@ -7,6 +7,14 @@
 
 #include "src/llv8-constants.h"
 
+using lldb::ByteOrder;
+
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+#define HOST_BYTE_ORDER ByteOrder::eByteOrderLittle
+#else
+#define HOST_BYTE_ORDER ByteOrder::eByteOrderBig
+#endif
```
bnoordhuis: I still think you should do this at runtime.  Both node and lldb do it that way.

hhellyer:
@bnoordhuis - I moved to a runtime check for endianess, do you want to review again?

Issue #100, opened 2017-06-19
workqueue command


brendangregg:
Would it make sense to have a "v8 workqueue" or "libuv workqueue" command, that showed what was waiting and their callbacks?

Maybe something that made use of  process._getActiveHandles(), process._getActiveRequests(), or process::getActiveHandles.

PR #122, opened 2017-08-17
src: add commands to inspect the workqueue

mmarchini:
Added two new commands (`v8 getactivehandles` and `v8 getactiverequests`) which prints all pending handles and requests (same return given by `process._getActiveHandles()` and `process._getActiveRequests()`). Those changes were built upon the symbols added on nodejs/node#14901, which means it's currently not working with node's latest build.  

Fixes: #100 
Ref: nodejs/node#14901
Ref: nodejs/post-mortem#46

### DIFF
--- a/src/llnode.cc
+++ b/src/llnode.cc
@@ -4,6 +4,8 @@
 #include <string.h>
 
 #include <cinttypes>
+#include <sstream>
+#include <string>
 
 #include <lldb/API/SBExpressionOptions.h>
 
@@ -11,6 +13,7 @@
 #include "src/llnode.h"
 #include "src/llscan.h"
 #include "src/llv8.h"
+#include "src/node-inl.h"
 
 namespace llnode {
 
@@ -20,6 +23,7 @@ using lldb::SBDebugger;
 using lldb::SBError;
 using lldb::SBExpressionOptions;
 using lldb::SBFrame;
+using lldb::SBProcess;
 using lldb::SBStream;
 using lldb::SBSymbol;
 using lldb::SBTarget;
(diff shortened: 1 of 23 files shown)


hhellyer:
Looks interesting, I can see there's some conversation going on over on nodejs/node. ;-)

Could you add some sample output to this PR? It's hard to test out your changes without building node as well.

A few thoughts (I'll do a proper review after we see what happens in nodejs/node!):
- If this is likely to be expanded on this later it might be worth moving it to it's own source file. That might be a good place for Node.js specific functionality.
- The constants are repeatedly loaded instead of being in llv8-constants.cc.
- You read pointers using process.ReadMemory instead of process.ReadPointerFromMemory which is simpler and correctly handles 32bit/64bit and endianess differences.
- Using ReadPointerFromMemory might also let you simplify your use of `buffer` and `myMemory` to be less confusing. (A more descriptive name for `myMemory` might be good too!)

src/llnode.cc
```diff
@@ -296,6 +306,184 @@ bool ListCmd::DoExecute(SBDebugger d, char** cmd,
...
+  int64_t envPtr = 0;
+  uint64_t env = 0;
+  int64_t queue = 0;
+  int64_t head = 0;
+  int64_t next = 0;
+  int64_t node = 0;
+  int64_t persistant_handle = 0;
+  v8::Error err2;
+
+  envPtr = LookupConstant(target, "nodedbg_currentEnvironment", envPtr, err2);
```
indutny: These should live in constants file.

indutny: Errors should be handled there too.

src/llnode.cc
```diff
@@ -296,6 +306,184 @@ bool ListCmd::DoExecute(SBDebugger d, char** cmd,
...
+  uint64_t buffer = 0;
+  bool go = true;
+  if (!thread.IsValid()) {
+    result.SetError("No valid process, please start something\n");
+    return false;
+  }
+
+  int activeHandles = 0;
+  uint64_t currentNode = env;
+  currentNode += queue;  // XXX env.handle_wrap_queue_
```
indutny: This really asks to be a class method.

src/llnode.cc
```diff
@@ -296,6 +306,184 @@ bool ListCmd::DoExecute(SBDebugger d, char** cmd,
...
+  currentNode += next;   // XXX env.handle_wrap_queue_.head_.next_
+  process.ReadMemory(currentNode, &buffer, size, sberr);
+  currentNode = buffer;
+  // TODO needs a stop condition, currently it's being stopped by a break
+  while (go) {
+    addr_t myMemory = currentNode;
+    myMemory = myMemory - node;  // wrap
+    myMemory += persistant_handle;
+    // XXX w->persistent().IsEmpty()
+    if (myMemory == 0) {
```
indutny: Class method too.

src/llnode.cc
```diff
@@ -296,6 +306,184 @@ bool ListCmd::DoExecute(SBDebugger d, char** cmd,
...
+    // TODO needs a better check
+    if (sberr.Fail()) {
+      break;
+    }
+
+    v8::JSObject v8_object(&llv8, buffer);
+    v8::Error err;
+    std::string res = v8_object.Inspect(&inspect_options, err);
+    if (err.Fail()) {
+      // result.SetError("Failed to evaluate expression");
```
indutny: No commented code is allowed.

src/llnode.cc
mmarchini: Maybe I can add these constants in a new `llnode-constants` file (instead of `llv8-constants`), as these symbols are related to node and not to v8?

mmarchini:
Thanks for the feedbacks! I'll do a full refactor after nodejs/node#14901 is merged, because probably there are going to be some changes which will impact this PR.

Here are the sample outputs of both commands:

screenshot from 2017-08-18 17-00-36

screenshot from 2017-08-18 17-00-45

src/llnode.cc
```diff
@@ -296,6 +306,184 @@ bool ListCmd::DoExecute(SBDebugger d, char** cmd,
...
+  llv8.Load(target);
+
+  int size = 8;
+  int64_t envPtr = 0;
+  uint64_t env = 0;
+  int64_t queue = 0;
+  int64_t head = 0;
+  int64_t next = 0;
+  int64_t node = 0;
+  int64_t persistant_handle = 0;
```
cjihrig: persistent is spelled incorrectly as persistant throughout.

mmarchini:
All code related to those new commands were refactored by assigning constants in a different file and by creating helper classes to access some node internal properties. I also added a "higher-level" `constants.h` file, which has classes that can be used by both `llnode-constants.h` and `llv8-constants.h` (I didn't change `llv8-constants.h` in this PR though to avoid making unrelated changes).

Besides that, a new fallback method to determine the current environment was added. This method uses a heuristic approach to try to find the root context of the process. With this, those new commands will work most of the time even without the new symbols added to V8.

There are probably still some error handling to be done. I'll do another review tomorrow and see what else can be improved.

Please tell me what you think and let me know if there are any changes necessary.

mmarchini:
Ok, since https://github.com/nodejs/node/pull/14901 landed and Node's postmortem metadata is now available on Node.js v9.7.0+ (also might be available on v8.11.0+ and v6.14.0+) I've refactored, added tests and rebased this PR. I believe it's good for review now! 😄 

P.S.: The PR grew a lot. If you want I can break it into two PRs (one to add Node's constants and other to add the new commands).

mmarchini:
Updated to fix an infinite loop in v6.x + missing constant for the same version: https://github.com/nodejs/llnode/pull/122/commits/63901e83f526a711c282430e43df088bd9c46351 and https://github.com/nodejs/llnode/pull/122/commits/3d2543f92b8ae19eb314be53986eb414c7a46e37

mmarchini:
Rebased to resolve some conflicts.

mmarchini:
Rebased to resolve conflicts again.

@indutny @hhellyer @cjihrig  would you like me to break this into 2 separate PRs?

mmarchini:
Superseded by #204 and #210

Issue #112, opened 2017-07-04
findjsinstances is not completely finished


gentlejo:
OS: RHEL7 
llnode version: 3.4.5

When I run findjsinstances command, it is not completely finished if its item is over about 100,000.

    (lldb) v8 findjsobjects
    ....
      24126    1338176 Object
      26602     854744 (Array)
     110907    2336544 (String)
    ....

    (lldb) v8 findjsinstances (String)
    killed (or Bus error)

If I upgrade to 3.9.x or 4.x, it shows better performance? Or what should I do to get result of big items?

bnoordhuis:
I suspect it's either some kind of memory corruption or that llnode simply runs out of memory or stack space.

Is there an easy way for me to reproduce, like a small test script I can run against a particular version of node?

hhellyer:
@gentlejo - I've landed a fix for external strings, could you verify by installing from the master version?
`npm install -g nodejs/llnode`
I think external strings are likely to have been your issue as well but it's possible v8 has another type of string we aren't handling properly! If you can verify the fix that would be great, if it's not fixed please re-open this.

gentlejo:
@hhellyer Oh so sorry for replying late. It was fixed! Thank you! 👍

PR #123, opened 2017-08-21
llv8-inl.h: Fix crash on sliced external strings

hhellyer:
Sliced strings with an external string as a parent would use their offset and length values against the string "(external)" as support for external strings is not implemented yet.

This may resolve issue #112 as I see a crash in the same way without this fix however a crash in the inspect code for another type would manifest in the same way. I will ask @gentlejo to check if this resolves the problem for them.

### DIFF
--- a/test/fixtures/inspect-scenario.js
+++ b/test/fixtures/inspect-scenario.js
@@ -38,6 +38,15 @@ function closure() {
   c.hashmap['internalized-string'] = 'foobar';
   // This thin string points to the previous 'foobar'.
   c.hashmap['thin-string'] = makeThin('foo', 'bar');
+  // Create an externalized string and slice it.
+  c.hashmap['externalized-string'] =
+    'string that will be externalized and sliced';
+  externalizeString(c.hashmap['externalized-string']);
+  // Sliced strings need to be longer that SlicedString::kMinLength
+  // which seems to be 13 so our string is 26.
+  c.hashmap['sliced-externalized-string'] =
+    c.hashmap['externalized-string'].substring(10,36);
+
   c.hashmap['array'] = [true, 1, undefined, null, 'test', Class];
   c.hashmap['long-array'] = new Array(20).fill(5);
   c.hashmap['array-buffer'] = new Uint8Array(
--- a/test/fixtures/inspect-scenario.js
+++ b/test/fixtures/inspect-scenario.js
@@ -40,12 +40,12 @@ function closure() {
   c.hashmap['thin-string'] = makeThin('foo', 'bar');
   // Create an externalized string and slice it.
   c.hashmap['externalized-string'] =
-    'string that will be externalized and sliced';
+      'string that will be externalized and sliced';
   externalizeString(c.hashmap['externalized-string']);
   // Sliced strings need to be longer that SlicedString::kMinLength
   // which seems to be 13 so our string is 26.
   c.hashmap['sliced-externalized-string'] =
-    c.hashmap['externalized-string'].substring(10,36);
+      c.hashmap['externalized-string'].substring(10,36);
 
   c.hashmap['array'] = [true, 1, undefined, null, 'test', Class];
   c.hashmap['long-array'] = new Array(20).fill(5);
--- a/test/inspect-test.js
+++ b/test/inspect-test.js
@@ -126,9 +126,9 @@ tape('v8 inspect', (t) => {
     ext = extMatch[1];
 
     const extSlicedMatch = lines.match(
-      /.sliced-externalized-string=(0x[0-9a-f]+):<String: "\(external\)">/);
+        /.sliced-externalized-string=(0x[0-9a-f]+):<String: "\(external\)">/);
     t.ok(extSlicedMatch,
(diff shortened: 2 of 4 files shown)


bnoordhuis:
It would be nice to have a regression test for this.  Starting node with `--expose_externalize_string` will let you create an external string with the `externalizeString('string')` global function.

hhellyer:
Will doing `externalizeString('string'). substring(...)` force creating a sliced string from an external string? (I'll try it but I don't know if V8 has rules around when it creates a sliced string versus creating a normal string.)

Shall I open an issue for external string support? It would be a nice self contained item if anyone wanted to pick it up.

hhellyer:
I've put in a test for creating an externalised string but I'm hitting a problem creating a sliced string from it that doesn't just create a non-external string.

The comments on SlicedString are making me suspicious that it's not possible:
https://github.com/nodejs/node/blob/master/deps/v8/src/objects.h#L7739
(Handling externalized strings is listed as a missing feature.)

I know llnode seemed to picked one up but it could have been in-flight data or a bug but it's also possible that comment is out of date.

bnoordhuis:
> I'm hitting a problem creating a sliced string from it that doesn't just create a non-external string.

Maybe I misunderstand you but isn't that the expected behavior?  Good point about that comment.  `NewSubString()` in factory.cc seems to handle external strings fine however:
```sh
$ node --allow_natives_syntax --expose_externalize_string
> externalizeString('boom')
undefined

> %SubString('boom', 0, 3)
'boo'

# and to prove it's externalized
> %ExternalStringGetChar('boom', 0)
98

# whereas a substring...
> %ExternalStringGetChar(%SubString('boom', 0, 3), 0)


#
# Fatal error in ../deps/v8/src/runtime/runtime-strings.cc, line 749
# Check failed: args[0]->IsExternalString().
#

==== C stack trace ===============================

    0   node                                0x0000000100bdd1b3 v8::base::debug::StackTrace::StackTrace() + 19
    1   node                                0x0000000100af09cb v8::platform::(anonymous namespace)::PrintStackTrace() + 27
    2   node                                0x0000000100bdb92c V8_Fatal + 220
    3   node                                0x00000001008f50c8 v8::internal::Runtime_ExternalStringGetChar(int, v8::internal::Object**, v8::internal::Isolate*) + 312
    4   ???                                 0x00000f8cff9188f8 0x0 + 17098257565944
Illegal instruction: 4
```

hhellyer:
It's confusing and I'm not sure if you are trying to do the exact same thing as me, so this is what I'm trying to add to the setup code for the testcase setup in inspect-scenario.js:

```
  c.hashmap['externalized-string'] =
    'string that will be externalized and sliced';
  externalizeString(c.hashmap['externalized-string']);
  c.hashmap['sliced-externalized-string'] =
    c.hashmap['externalized-string'].substring(0,6);
```

The problem is that in the resulting dump the object `sliced-externalized-string` refers to either isn't a sliced string or doesn't have an external string as a parent, I'm not sure which yet. I'll look at this some more when I get back from holiday. (I'm out next week.)

hhellyer:
I've added the test case, the problem was that to get a sliced string the substring needs to be longer that `SlicedString::kMinLength`. That's set to 13 so now I'm creating a longer substring and the `SlicedString` does reference the `ExternalString`.

test/fixtures/inspect-scenario.js
```diff
@@ -38,6 +38,15 @@ function closure() {
...
   // This thin string points to the previous 'foobar'.
   c.hashmap['thin-string'] = makeThin('foo', 'bar');
+  // Create an externalized string and slice it.
+  c.hashmap['externalized-string'] =
+    'string that will be externalized and sliced';
+  externalizeString(c.hashmap['externalized-string']);
+  // Sliced strings need to be longer that SlicedString::kMinLength
+  // which seems to be 13 so our string is 26.
+  c.hashmap['sliced-externalized-string'] =
+    c.hashmap['externalized-string'].substring(10,36);
```
bnoordhuis: Four space indent for line continuations.

test/inspect-test.js
```diff
@@ -118,6 +120,17 @@ tape('v8 inspect', (t) => {
     t.ok(thinMatch, '.thin-string ThinString property');
     thin = thinMatch[1];
 
+    const extMatch = lines.match(
+        /.externalized-string=(0x[0-9a-f]+):<String: "\(external\)">/);
+    t.ok(extMatch, '.externalized-string ExternalString property');
+    ext = extMatch[1];
+
+    const extSlicedMatch = lines.match(
+      /.sliced-externalized-string=(0x[0-9a-f]+):<String: "\(external\)">/);
```
bnoordhuis: Ditto.

test/inspect-test.js
```diff
@@ -118,6 +120,17 @@ tape('v8 inspect', (t) => {
...
+    t.ok(extSlicedMatch,
+      '.sliced-externalized-string Sliced ExternalString property');
```
bnoordhuis: Can you line up the arguments?

PR #125, opened 2017-08-21
llscan.cc: Add findjsinstances --verbose option

hhellyer:
This adds a `--inspect`, `-i` option to findjsinstances so that the output of `v8 inspect` is displayed for each instance of a type instead of `v8 print` output. This makes it quicker to view the contents of objects especially where there are few instances of the same kind.

If there's only one instance of `MyObject` then it's easier to type:
```
(llnode) v8 findjsinstances -i MyObject
```
than
```
(llnode) v8 findjsinstances MyObject
0x00002d2fae211241:<Object: MyObject>
(llnode) v8 inspect 0x00002d2fae211241
```
if there's a small number then it's much easier than cut and pasting each hexadecimal reference in turn.

I chose `--inspect` as that matches `v8 inspect` however it could also be `--verbose` and `-v`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,31 @@ make -C out/ -j9
 sudo make install-linux
 ```
 
+### FreeBSD
+
+```bash
+# Clone this repo
+git clone https://github.com/nodejs/llnode.git && cd llnode
+
+# Install llvm with lldb and headers
+pkg install llvm39
+rm -f /usr/bin/lldb
+ln -s /usr/local/bin/lldb39 /usr/bin/lldb
+
+# Initialize GYP
+git clone https://chromium.googlesource.com/external/gyp.git tools/gyp
+
+# Configure
+./gyp_llnode -Dlldb_dir=/usr/local/llvm39/
+
+# Build
+gmake -C out/ -j9
+```
+
+(The LLDB function ComputeSystemPluginsDirectory is not implemented on FreeBSD.
+The plugin library must be loaded manually.)
+
+
 ## Loading the lldb plugin library.
 
 The simplest method is:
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -224,7 +224,8 @@ The following subcommands are supported:
                          dumped.
 
                          Syntax: v8 bt [number]
-      findjsinstances -- List all objects which share the specified map.
+      findjsinstances -- List every object with the specified type name.
+                         Use -i or --inspect to display detailed `v8 inspect` outputfor each object.
```
cjihrig: outputfor -> output for

hhellyer: Good spot, it was in the help string in llnode.cc so I've corrected both places.

mmarchini:
Inspecting the `process` object will be so much easier with this!

hhellyer:
@mmarchini - It should help with things like the `process` object where you know ahead of time there's only one *but* in that specific case we do have `v8 nodeinfo` to produce a nice human readable summary of it's contents. (Because it's always good to be able to check you are looking at the dump you meant to look at.)

src/llscan.cc
```diff
@@ -101,7 +101,20 @@ bool FindInstancesCmd::DoExecute(SBDebugger d, char** cmd,
...
+    if (strcmp(*start, "-i") == 0 || strcmp(*start, "--inspect") == 0) {
+      inspect_options.detailed = true;
+      // Shuffle up the remaining parameters.
+      char** curr = start;
+      for (; *curr != nullptr; curr++) {
+        char** next = curr + 1;
+        *curr = *next;
+      }
+    }
+  }
```
bnoordhuis: IMO, this ad hoc option parser is kind of nasty.  Is there a reason you can't extend `ParseInspectOptions()`?

geek:
instead of `--inspect` any thoughts on calling this `--verbose`,`-v` ... the option that mdb_v8 uses.

hhellyer:
@geek - I did mention that in the first comment. My reasoning was that it causes findjsinstances to execute the inspect command for each object. I don't have a strong preference either way though and I can probably change string constants without invalidating the approval.

Does anyone else have an opinion?

cjihrig:
I think keeping the terminology similar between llnode and mdb_v8 would be beneficial, especially for people coming here from mdb, or people who use both.

PR #130, opened 2017-09-01
src,test: support V8 6.1 postmortem data

cjihrig:
This is still a work in progress, but adds postmortem debugging support for V8 6.1. The following issues still need to be resolved (all marked with `TODO` in the code):

- [x] ~~Properly handle the end offset of function bodies. It previously needed to be shifted, but that is no longer the case. This just needs version detection logic.~~
- [x] ~~Properly load the `ThinStringTag` constant. `V8DBG_THINSTRINGTAG` appears to be defined, but the constant isn't loading. Right now, the value of `0x5` is hard coded.~~
- [x] ~~Support `Uint8Array`s. Those tests are currently broken and commented out. The problem is that the backing store of the `JSArrayBufferView`'s `JSArrayBuffer` is coming back as all zeros.~~

UPDATE: The issue with the `Uint8Array`s is that the backing stores are not materialized. If the test is updated to reference the `.buffer` property, the backing store is materialized and shows up properly in llnode.

### DIFF
--- a/src/llv8-constants.cc
+++ b/src/llv8-constants.cc
@@ -471,7 +471,6 @@ void DescriptorArray::Load() {
         LoadConstant("prop_attributes_DONT_ENUM");
 
     kPropertyKindMask = LoadConstant("prop_kind_mask");
-    kPropertyKindShift = LoadConstant("prop_kind_shift", int64_t(0));
     kPropertyKindEnum_kAccessor = LoadConstant("prop_kind_Accessor");
     kPropertyKindEnum_kData = LoadConstant("prop_kind_Data");
 
--- a/src/llv8-constants.cc
+++ b/src/llv8-constants.cc
@@ -212,8 +212,13 @@ void Map::Load() {
 
 
 void JSObject::Load() {
-  kPropertiesOffset = LoadConstant("class_JSReceiver__properties__FixedArray",
-                                   "class_JSObject__properties__FixedArray");
+  kPropertiesOffset =
+      LoadConstant("class_JSReceiver__raw_properties_or_hash__Object",
+                   "class_JSReceiver__properties__FixedArray");
+
+  if (kPropertiesOffset == -1)
+    kPropertiesOffset = LoadConstant("class_JSObject__properties__FixedArray");
+
   kElementsOffset = LoadConstant("class_JSObject__elements__Object");
   kInternalFieldsOffset =
       LoadConstant("class_JSObject__internal_fields__uintptr_t");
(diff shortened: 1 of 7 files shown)


src/llv8-constants.cc
```diff
@@ -334,7 +346,8 @@ void String::Load() {
   kConsStringTag = LoadConstant("ConsStringTag");
   kSlicedStringTag = LoadConstant("SlicedStringTag");
   kExternalStringTag = LoadConstant("ExternalStringTag");
-  kThinStringTag = LoadConstant("ThinStringTag");
+  // TODO: The 0x5 on the following line shouldn't be needed.
+  kThinStringTag = LoadConstant("ThinStringTag", 0x5);
```
indutny: For what node.js/v8 version `0x5` is needed?

src/llv8-inl.h
```diff
@@ -259,7 +259,9 @@ inline int64_t SharedFunctionInfo::EndPosition(Error& err) {
   if (err.Fail()) return -1;
 
   field &= 0xffffffff;
-  return field >> 1;
+  // TODO: Offset should be zero in new versions, but it breaks old versions.
+  // return field >> 1;
+  return field;
```
indutny: Isn't this going to break compatibility with old versions?

indutny:
Yikes, work in progress. Sorry for premature review!

cjihrig:
No worries, but yes, your two comments correspond to a couple of the remaining problems :-)

cjihrig:
@indutny thanks for the help on IRC. I was able to get all of the tests passing on V8 6.1. Apparently, the backing stores of typed arrays are not necessarily materialized like they were in older versions of V8. If you update the inspect-scenario to simply access `c.hashmap['uint8-array'].buffer`, then it is materialized and works properly in llnode.

Currently, I'm reporting unmaterialized backing stores as being unmaterialized. I think I can materialize the backing store in llnode, but it will take more work.

cjihrig:
This is ready for review.

I've done regression testing locally, and all tests are passing with Node 4.8.4 (latest Argon), 6.11.2 (latest Boron), 8.2.1 (latest Node 8 with Crankshaft).

I also tested [the V8 6.1 branch](https://github.com/cjihrig/node-1/tree/v8-6.1), with [these changes to the postmortem metadata](https://github.com/cjihrig/node-1/commit/92690a4abe8a448587b9f08fc03df12f78120182).

src/llv8-constants.cc
```diff
@@ -537,7 +562,8 @@ void Types::Load() {
   kOddballType = LoadConstant("type_Oddball__ODDBALL_TYPE");
   kJSObjectType = LoadConstant("type_JSObject__JS_OBJECT_TYPE");
   kJSAPIObjectType = LoadConstant("APIObjectType");
-  kJSSpecialAPIObjectType = LoadConstant("APISpecialObjectType");
+  kJSSpecialAPIObjectType = LoadConstant("SpecialAPIObjectType",
+                                         "APISpecialObjectType");
```
bnoordhuis: I think this has always been called `SpecialAPIObjectType`.  `APISpecialObjectType` may have been a typo.

@indutny Can you confirm?

src/llv8-inl.h
```diff
@@ -259,7 +259,8 @@ inline int64_t SharedFunctionInfo::EndPosition(Error& err) {
   if (err.Fail()) return -1;
 
   field &= 0xffffffff;
-  return field >> 1;
+  field >>= v8()->shared_info()-> kEndPositionShift;
```
bnoordhuis: Space after `->`.

indutny:
`make format` is your friend. Will take a deeper look after that :wink: Thanks for the changes!

src/llv8-constants.cc
```diff
@@ -262,19 +267,27 @@ void JSDate::Load() {
...
+                             "class_SharedFunctionInfo__name__Object");
   kInferredNameOffset =
       LoadConstant("class_SharedFunctionInfo__inferred_name__String",
                    "class_SharedFunctionInfo__function_identifier__Object");
   kScriptOffset = LoadConstant("class_SharedFunctionInfo__script__Object");
   kCodeOffset = LoadConstant("class_SharedFunctionInfo__code__Code");
   kStartPositionOffset =
-      LoadConstant("class_SharedFunctionInfo__start_position_and_type__SMI");
+      LoadConstant("class_SharedFunctionInfo__start_position_and_type__int",
+                   "class_SharedFunctionInfo__start_position_and_type__SMI");
```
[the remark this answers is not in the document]
cjihrig: @bnoordhuis it looks like those values are already exposed as `v8dbg_sharedfunctioninfo_start_position_mask` and `v8dbg_sharedfunctioninfo_start_position_shift`. It looks like llnode is loading them, and if the constants aren't found, falling back to those hard coded values.

cjihrig:
@indutny I ran `make format`. Sorry I missed that. It might be good to incorporate that as part of `npm test`.

hhellyer:
@cjihrig - Would you like me to merge this now or wait until 6.1 is landed in Node.js? (We've a few approved PR's on llnode atm. They do seem to merge cleanly when I test locally.)

cjihrig:
@hhellyer it should be fine to land this before 6.1 lands in core.

Issue #133, opened 2017-09-05
llnode doesn't support lldb 6.0 now


langhuihui:
function getLinuxVersion(lldbExe) {

  var lldbStr;
  try {
    lldbStr = child_process.execFileSync(lldbExe, ['-v']).toString();
  } catch (err) {
    return undefined;
  }
  // Ignore minor revisions like 3.8.1
  if (lldbStr.indexOf('version 4.0') > 0) {
    return '4.0';
  } else if (lldbStr.indexOf('version 3.9') > 0) {
    return '3.9';
  } else if (lldbStr.indexOf('version 3.8') > 0) {
    return '3.8';
  } else if (lldbStr.indexOf('version 3.7') > 0) {
    return '3.7';
  } if (lldbStr.indexOf('version 3.6') > 0) {
    return '3.6';
  } if (lldbStr.indexOf('version 3.5') > 0) {
    return '3.5';
  }
  return undefined;
}

need improve the code

bnoordhuis:
There is no lldb 6.0, the 4.0 series are the latest at this time.

hhellyer:
v5.0 is due for release soon and the release_50 branch now exists in git, we should fix this - it's useful for anyone who compiles lldb themselves. (Like me!)

I'll try to do a PR later on.

langhuihui:
@hhellyer yes. I am sure that lldb version shows 6.0 when I  compiles lldb.

bnoordhuis:
PR welcome but the point remains that 4.0 is the latest release and this bug report is invalid.  I think it's perfectly sane for the install script to fail on prerelease versions of lldb.

@langhuihui You are using lldb from git master.  I wouldn't recommend that in conjunction with llnode - we don't test that and can't guarantee it works.  Use a release version.

hhellyer:
I won't re-open again as otherwise this will turn into a game of ping-pong but:
* If we hit problems opening particular core dumps we would recommend users try the latest version of lldb with llnode as that's where any fix would arrive first. (Some of those fixes have been from us.)
* lldb-5.0 can be installed from http://apt.llvm.org/ now. Users can get it there before it hits the main repositories both pre-release and once it's formally released.
* The lldb team make a big effort to keep the C++ API compatible across releases and enforce rules when adding to it that ensure it works. (I'd normally be wary of this but the wider llvm project does produce a C++ compiler so I suspect they know what they are are doing.)

PR #134, opened 2017-09-05
configure.js: Simplify lldb version handling

hhellyer:
This change removes the hardcoded map of lldb versions to release branches. This makes it easier to use the --lldb_exe= parameter to handle unknown releases should find the correct branch for pre-release lldb builds or local builds of lldb.
It also makes the the installer aware of lldb-5.0 binaries on Linux since pre-release versions can be installed directly from llvm.org, see: http://apt.llvm.org/

This fixes issue #133

### DIFF
--- a/scripts/configure.js
+++ b/scripts/configure.js
@@ -5,16 +5,6 @@ const fs = require('fs');
 const path = require('path');
 const child_process = require('child_process');
 
-const lldbReleases = {
-  '4.0': 'release_40',
-  '3.9': 'release_39',
-  '3.8': 'release_38',
-  '3.7': 'release_37',
-  '3.6': 'release_36',
-  '3.5': 'release_35',
-  '3.4': 'release_34',
-};
-
 const buildDir = process.cwd();
 
 console.log('Build dir is: ' + buildDir);
@@ -39,7 +29,7 @@ if (osName === 'Darwin') {
     process.exit(1);
   }
 
-  lldbHeadersBranch = lldbReleases[lldbVersion];
+  lldbHeadersBranch = lldbVersionToBranch(lldbVersion);
   lldbIncludeDir = 'lldb-' + lldbVersion;
 
 } else if (osName === 'Linux') {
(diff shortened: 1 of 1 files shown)


hhellyer:
@langhuihui - would you be able to do:
`npm install -g hhellyer/llnode#lldb_versions`
to verify this fixes issue #133 for you? 

You might need to do:
`npm install -g hhellyer/llnode#lldb_versions --lldb_exe=<path_to_lldb_build>/bin/lldb`
to use your local build if it's not on your path.

langhuihui:
thank u very much.God bless you

scripts/configure.js
```diff
@@ -205,18 +200,9 @@ function getLinuxVersion(lldbExe) {
...
-    return '3.8';
-  } else if (lldbStr.indexOf('version 3.7') > 0) {
-    return '3.7';
-  } if (lldbStr.indexOf('version 3.6') > 0) {
-    return '3.6';
-  } if (lldbStr.indexOf('version 3.5') > 0) {
-    return '3.5';
+  let versionMatch = lldbStr.match(/version (\d.\d)/);
+  if (versionMatch) {
+      return versionMatch[1];
```
bnoordhuis: Indent error.

scripts/configure.js
```diff
@@ -140,6 +130,10 @@ fs.writeFileSync(`${buildDir}/scripts/llnode.sh`, scriptText(lldbExe));
 // Exit with success.
 process.exit(0);
 
+function lldbVersionToBranch(version) {
+    return 'release_' + version.replace('.','');
```
bnoordhuis: Indent error and space after comma.

hhellyer:
@bnoordhuis - I've fixed the white space. I'm not sure about adding the warnings, keeping the supported versions in sync is likely to be fragile and cause people to see spurious warnings and raise bugs. I'm not sure anyone is likely to use a pre-release version of lldb by accident - you either have to add a package repository (as root) or compile it from source to get it onto your system.

Issue #138, opened 2017-09-30
v8 source list always fails w/ error: USAGE: v8 source list


No9:
I upgraded llnode install 1.6.2 on both Ubuntu 16.04 and FreeBSD 12 and ran through the tutorial here 
https://developer.ibm.com/node/2016/08/15/exploring-node-js-core-dumps-using-the-llnode-plugin-for-lldb/ 

When a frame is selected (e.g. `frame select 6`) and the source viewed with `v8 source list` the response is always 
`error: USAGE: v8 source list`

@cjihrig This seems to have been introduced in 1.6.0 
https://github.com/nodejs/llnode/pull/106/commits/1d43035f9134201f37274b63cc6efdd4672010d1#diff-81649c6431efceafdaca2ec7afb55f38R187 

When the `*cmd == nullptr` part of the conditional is removed the source is printed as expected.

I can PR the removal but would appreciate a second pair of eyes to confirm it won't impact the original seg faults fix.

Thanks

cjihrig:
That seems fine, just verify that `jssource` still works for you too.

FWIW, the command seems broken on my machine, even with that change. I'll try to take a deeper look when I have a chance.

jamischarles:
I ran into this issue too. My workaround was the following given this frame:

```
frame #6: 0x00002e5b6a392f12 missingParamFunc(this=0x00002e41b0f20031:<Object: Object>, 0x00003ca8de102241:<undefined>) at (no script):8:26 fn=0x00002e41b0f1ffb1
```

1) `select frame 6`
2) 
```
(lldb) v8 source list 0x00002e41b0f1ffb1
  8 function missingParamFunc(param1) {
  9 	console.log('test', param1.bar)
  10 }
```

Sometimes it errors out with this `error: 'source list' takes no arguments, only flags.`, but it seems to work with the above method when I run it immediately after selecting a frame.

robertkowalski:
same here, so far i haven't been able to print any source

misterdjules:
@robertkowalski The issue you're running into seems like it might be a different problem. Would you mind opening a separate issue that provide as much detail as possible, such as the node version you're using, and detailed repro steps?

PR #139, opened 2017-10-01
Fix for printing source

No9:
This update fixes https://github.com/nodejs/llnode/issues/138 
Tested on Ubuntu 16.04 and FreeBSD 12 
As per the feedback `jssource` has been tested and also works


joyeecheung:
@No9 Can you add a test in `test/inspect-test.js`? You can grab the frame where `method` is at in `line.match(/method\(this=(0x[0-9a-f]+)[^\n]+fn=(0x[0-9a-f]+)/i);` and add a test case running `frame select` then inspect the source there.

mmarchini:
Fixed in https://github.com/nodejs/llnode/pull/259

PR #140, opened 2017-10-06
doc: update LLNODE_RANGESFILE documentation

bnoordhuis:
Document that the LLNODE_RANGESFILE environment variable is not needed
with lldb >= 3.9.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,8 +228,8 @@ The following subcommands are supported:
                          Use -v or --verbose to display detailed `v8 inspect` output for each object.
                          Accepts the same options as `v8 inspect`
       findjsobjects   -- List all object types and instance counts grouped by typename and sorted by instance count.
-                         Requires `LLNODE_RANGESFILE` environment variable to be set to a file containing memory ranges for the
-                         core file being debugged.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -228,8 +228,8 @@ The following subcommands are supported:
                          Use -v or --verbose to display detailed `v8 inspect` output for each object.
                          Accepts the same options as `v8 inspect`
       findjsobjects   -- List all object types and instance counts grouped by typename and sorted by instance count.
-                         Requires `LLNODE_RANGESFILE` environment variable to be set to a file containing memory ranges for the
```
indutny: Should we also update the source file that generated this?

bnoordhuis: This line is wrapped in an #ifdef guard in the source, i.e., it doesn't print it if lldb >= 3.9.

rnchamberlain:
LGTM. 

Comment: with the #ifdef in the source, cut&pasting into the readme did not really work. Also, on the NPM landing page it is hard to read this information as you have to scroll left/right on a long chunk of text. Maybe it would be better re-written and formatted in the readme as a nice list of commands.

indutny:
Verbal LGTM

joyeecheung:
This should be good to merge now.

PR #141, opened 2017-10-21
Try to use the version from llvm-config on MacOS

joyeecheung:
Previously if one has `lldb` executable set to something different from what `xcodebuild` integrates, since the plugin will be linked to the default lldb instead of the one loading it, we will get an error when running any llnode commands:

```
error: No valid process, please start something
```

This PR fixes that by detecting the lldb libraries from `llvm-config` first and use that if available. The theory is, if one has a `lldb` different from the default one, then they probably have `llvm-config` setup as well (e.g. `brew install llvm --with-lldb` and configure the PATH to use that version).

Also skip cloning the headers if they are already cloned.

### DIFF
--- a/scripts/configure.js
+++ b/scripts/configure.js
@@ -82,12 +82,19 @@ console.log(`Installing llnode for ${lldbExe}, lldb version ${lldbVersion}`);
 // TODO: The llvm project is probably moving to github soon at that point we
 // should stop using the mirror.
 if (lldbHeadersBranch != undefined) {
-  console.log('Cloning lldb from ' + lldbHeadersBranch);
-  child_process.execSync(`rm -rf ${lldbIncludeDir}`);
-  child_process.execFileSync('git',
+  if (!fs.lstatSync(path.join(lldbIncludeDir, 'include')).isDirectory()) {    
+    console.log('Cloning lldb from ' + lldbHeadersBranch);  
+    child_process.execSync(`rm -rf ${lldbIncludeDir}`);
+    child_process.execFileSync('git',
     ['clone', '--depth=1', '-b', lldbHeadersBranch,
       'https://github.com/llvm-mirror/lldb.git', lldbIncludeDir],
-    {cwd: buildDir});
+    {
+      cwd: buildDir,
+      stdio: 'inherit'  // show progress
+    });
+  } else {
+    console.log(`Skip cloning lldb headers because ${lldbIncludeDir} exists`);
+  }
 }
 
 // Link to the headers file so we can run gyp_llnode directly and don't need to
@@ -136,6 +143,20 @@ function lldbVersionToBranch(version) {
 
 // On Mac the lldb version string doesn't match the original lldb versions.
 function getDarwinRelease() {
+  try {
+    const version = child_process.execFileSync('llvm-config', ['--version']).toString().trim();
+    const prefix = child_process.execFileSync('llvm-config', ['--prefix']).toString().trim();
+    const options = JSON.stringify({
+      variables: { "lldb_build_dir%": prefix }
+    }, null, 2);
+    fs.writeFileSync('options.gypi', Buffer.from(options), 'utf-8');
+    console.log('Overwriting options.gypi with output from llvm-config:');
+    console.log(options);
+    return version.split('.').slice(0, 2).join('.');
+  } catch(err) {
+    // No llvm-config, try to get the version from xcodebuild
+  }
+
   var xcodeStr;
   try {
     xcodeStr = child_process.execFileSync('xcodebuild', ['-version'])
--- a/scripts/configure.js
+++ b/scripts/configure.js
@@ -28,7 +28,7 @@ if (osName === 'Darwin') {
   }
 
   console.log(`Installing llnode for ${lldbExe}, lldb version ${lldbVersion}`);
-  const installedDir = getDarwinIntallDir();
+  const installedDir = getDarwinInstallDir();
   if (installedDir === undefined) {
     const lldbHeadersBranch = lldbVersionToBranch(lldbVersion);
     lldbInstallDir = 'lldb-' + lldbVersion;
@@ -90,8 +90,7 @@ function getLldbHeadersPath(lldbInstallDir) {
 // should stop using the mirror.
 function cloneHeaders(lldbHeadersBranch, lldbInstallDir, buildDir) {
   const lldbHeaders = getLldbHeadersPath(lldbInstallDir);
-  if (!fs.existsSync(lldbHeaders)) {
-    child_process.execSync(`rm -rf ${lldbInstallDir}`);
+  if (!fs.existsSync(lldbInstallDir)) {
     console.log(`Cloning lldb from ${lldbHeadersBranch} to ${lldbInstallDir}`);
     child_process.execFileSync('git',
       ['clone', '--depth=1', '-b', lldbHeadersBranch,
@@ -174,7 +173,7 @@ function getDarwinRelease() {
   var versionStr = '';
   var splitStr = xcodeStr.split(os.EOL);
   for (var str of splitStr) {
-    if (str.indexOf('Xcode') !== -1) {
+    if (str.includes('Xcode')) {
       versionStr = str.split(' ')[1];
       break;
     }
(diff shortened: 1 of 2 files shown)


joyeecheung:
The `options.gypi` include is already in `gyp_llnode`. I am not sure it is safe to overwrite the whole thing but there does not seem to be any usage of `options.gypi` previously in the codebase.

scripts/configure.js
```diff
@@ -82,12 +82,19 @@ console.log(`Installing llnode for ${lldbExe}, lldb version ${lldbVersion}`);
 // TODO: The llvm project is probably moving to github soon at that point we
 // should stop using the mirror.
 if (lldbHeadersBranch != undefined) {
-  console.log('Cloning lldb from ' + lldbHeadersBranch);
-  child_process.execSync(`rm -rf ${lldbIncludeDir}`);
-  child_process.execFileSync('git',
+  if (!fs.lstatSync(path.join(lldbIncludeDir, 'include')).isDirectory()) {
```
bnoordhuis: `lstatSync()` throws if the path doesn't exist.

scripts/configure.js
```diff
@@ -82,12 +82,19 @@ console.log(`Installing llnode for ${lldbExe}, lldb version ${lldbVersion}`);
 // TODO: The llvm project is probably moving to github soon at that point we
 // should stop using the mirror.
 if (lldbHeadersBranch != undefined) {
-  console.log('Cloning lldb from ' + lldbHeadersBranch);
-  child_process.execSync(`rm -rf ${lldbIncludeDir}`);
-  child_process.execFileSync('git',
+  if (!fs.lstatSync(path.join(lldbIncludeDir, 'include')).isDirectory()) {    
+    console.log('Cloning lldb from ' + lldbHeadersBranch);  
+    child_process.execSync(`rm -rf ${lldbIncludeDir}`);
```
bnoordhuis: I know it's not new but seeing `rm -rf` in an install script always gives me the creeps...

scripts/configure.js
```diff
@@ -136,6 +143,20 @@ function lldbVersionToBranch(version) {
 
 // On Mac the lldb version string doesn't match the original lldb versions.
 function getDarwinRelease() {
+  try {
+    const version = child_process.execFileSync('llvm-config', ['--version']).toString().trim();
+    const prefix = child_process.execFileSync('llvm-config', ['--prefix']).toString().trim();
```
bnoordhuis: Long lines.

scripts/configure.js
```diff
@@ -136,6 +143,20 @@ function lldbVersionToBranch(version) {
...
+    const options = JSON.stringify({
+      variables: { "lldb_build_dir%": prefix }
+    }, null, 2);
+    fs.writeFileSync('options.gypi', Buffer.from(options), 'utf-8');
```
bnoordhuis: Using `Buffer.from()` means this script will fail for people stuck on node.js versions < v4.5.0.  It also doesn't look necessary to me or am I missing something?

scripts/configure.js
```diff
@@ -136,6 +143,20 @@ function lldbVersionToBranch(version) {
...
+    console.log('Overwriting options.gypi with output from llvm-config:');
+    console.log(options);
+    return version.split('.').slice(0, 2).join('.');
+  } catch(err) {
```
bnoordhuis: This catch block should arguably be just for the `.execFileSync()` calls.  Rule of thumb: catch blocks should be no bigger than absolutely necessary, particularly catch-and-swallow blocks.

scripts/configure.js
joyeecheung: Yea I thought about removing this...I guess we can just clone if the directory is not there, otherwise do nothing?

scripts/configure.js
joyeecheung: You're right, we don't need to transform this into a buffer

scripts/configure.js
joyeecheung: Good catch. Will fix that, thanks

scripts/configure.js
joyeecheung: Thanks for catching that (no pun intended)

scripts/configure.js
indutny: Or use `rimraf`

joyeecheung:
@indutny @bnoordhuis Thanks for the reviews, I will address them later. Also this just came to my mind: we don't need to clone the headers if we can get them using `llvm-config --includedir`. Let me know if I missed something.

scripts/configure.js
joyeecheung: @indutny `script/configure.js` is run during preinstall so we can't require `rimraf` at that point..

joyeecheung:
@bnoordhuis @indutny Updated, PTAL. I also cleaned up the script a little bit (white spaces, etc) with a quick eslint config

```
{
  "extends": "standard",
  "rules": {
    "semi": ["error", "always"],
    "space-before-function-paren": ["error", "never"],
    "no-multi-spaces": ["error", { "ignoreEOLComments": true }],
    "camelcase": "off"
  }
}
```

This now skip downloading headers when <code>\`llvm-config --prefix\`/include/lldb</code> exists. 
Also I have renamed the `lldbIncludeDir` to `lldbInstallDir`, because AFAICT this variable refers to what `llvm-config --prefix` (`Print the installation prefix` as the help text describe) should point to, so naming it `lldbIncludeDir` is a bit confusing because the actual directory to include is that + `/include`(`llvm-config --includedir`)

scripts/configure.js
```diff
@@ -62,43 +66,54 @@ if (osName === 'Darwin') {
...
-  console.log('Cloning lldb from ' + lldbHeadersBranch);
-  child_process.execSync(`rm -rf ${lldbIncludeDir}`);
-  child_process.execFileSync('git',
-    ['clone', '--depth=1', '-b', lldbHeadersBranch,
-      'https://github.com/llvm-mirror/lldb.git', lldbIncludeDir],
-    {cwd: buildDir});
+function cloneHeaders(lldbHeadersBranch, lldbInstallDir, buildDir) {
+  const lldbHeaders = getLldbHeadersPath(lldbInstallDir);
+  if (!fs.existsSync(lldbHeaders)) {
+    child_process.execSync(`rm -rf ${lldbInstallDir}`);
```
bnoordhuis: Looking at this again, won't this fail horribly if `lldbInstallDir` contains spaces?

scripts/configure.js
```diff
@@ -11,29 +11,34 @@ console.log('Build dir is: ' + buildDir);
...
   if (lldbVersion === undefined) {
     console.log('Unable to locate lldb binary. llnode installation failed.');
     process.exit(1);
   }
 
-  lldbHeadersBranch = lldbVersionToBranch(lldbVersion);
-  lldbIncludeDir = 'lldb-' + lldbVersion;
-
+  console.log(`Installing llnode for ${lldbExe}, lldb version ${lldbVersion}`);
+  const installedDir = getDarwinIntallDir();
```
bnoordhuis: Typo: `getDarwinInstallDir()` (missing `s`)

scripts/configure.js
```diff
@@ -136,17 +151,30 @@ function lldbVersionToBranch(version) {
...
+      '-version'
+    ]).toString();
   } catch (err) {
     return undefined;
   }
   var versionStr = '';
   var splitStr = xcodeStr.split(os.EOL);
   for (var str of splitStr) {
-    if (str.indexOf('Xcode') != -1) {
+    if (str.indexOf('Xcode') !== -1) {
```
bnoordhuis: You could use `str.includes('Xcode')` here.

scripts/configure.js
joyeecheung: @bnoordhuis Ouch..I think maybe we can just skip deleting the directory altogether. In case there are any stale links, the user can totally remove them by hand..

joyeecheung:
@bnoordhuis @indutny Updated, PTAL, thanks

joyeecheung:
Ping @indutny do you still request changes? They should be addressed now.

joyeecheung:
Ping @indutny again, though from the text of that review

> All that @bnoordhuis said, plus a suggestion.

I guess the red cross could be dismissed now.

bnoordhuis:
I've taken the liberty of dismissing Fedor's review since his comments (well, singular) were addressed.

joyeecheung:
@bnoordhuis Thanks, I have merged it so `npm test` works with custom lldb on mac again.

PR #142, opened 2017-10-30
Update travis badge.

cPhost:


joyeecheung:
On a side now, what's the status with the Travis CI? At the moment all builds fail with `test timed out after 15000ms`, looks like lldb printed some errors on the test commands. Or are we moving to Jenkins? (https://github.com/nodejs/build/issues/777)

cPhost:
@joyeecheung `tape test/*-test.js` command is timeout for some reason.

joyeecheung:
@cPhost Yes it times out when the expected output does not come out from the lldb session, most of the time it's because lldb got into some errors so instead of the expected output, it writes the error message. We should probably improve the tests by setting a timeout ourselves and log the session output when it times out.

PR #144, opened 2017-11-03
test: update travis config, handle errors properly, add TEST_LLNODE_DEBUG

joyeecheung:
At the moment when a test times out (usually because the expected lldb output does not show up, possible due to an error), there is only a tape timeout message, as is what's happening in our Travis CI build.

This PR checks the timeout and throws an error containing the lldb output for the ease of debugging.

I've tried it on Mac and Linux with v8.9.0 and v6.11.5. I've noticed that the tests are a bit flaky in master, with random timeouts (with this patch on I can see the sent command got swallowed, it's `8 <command>` instead of `v8 <command>` so lldb will error and the test times out). Sometimes there is a ECONNRESET on the pipe. I'll investigate that later.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,6 @@ plugin: configure
 	$(MAKE) -C out/
 
 _travis: plugin
-	TEST_LLDB_BINARY=`which lldb-3.9` npm test
+	TEST_LLDB_BINARY=`which lldb-3.9` TEST_LLNODE_DEBUG=true npm test
 
 .PHONY: all
--- a/Makefile
+++ b/Makefile
@@ -19,9 +19,14 @@ uninstall-linux:
 format:
 	clang-format -i src/*
 
-_travis:
-	./gyp_llnode -Dlldb_dir=/usr/lib/llvm-3.6/ -f make
-	make -C out/
-	TEST_LLDB_BINARY=`which lldb-3.6` npm test
+configure: scripts/configure.js
+	node scripts/configure.js
+
+plugin: configure
+	./gyp_llnode
+	$(MAKE) -C out/
+
+_travis: plugin
+	TEST_LLDB_BINARY=`which lldb-3.9` npm test
 
 .PHONY: all
--- a/test/common.js
+++ b/test/common.js
@@ -62,7 +62,9 @@ function SessionOutput(session, stream, timeout) {
   });
 
   // Ignore errors
-  stream.on('error', () => {});
+  stream.on('error', (err) => {
+    debug('[stream error]', err);
+  });
 }
 util.inherits(SessionOutput, EventEmitter);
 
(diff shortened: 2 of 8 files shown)


joyeecheung:
The Travis builds now error instead of timing out. There is a reproduction of the "swallowed v": https://travis-ci.org/nodejs/llnode/jobs/296685259

joyeecheung:
Just noticed that since we only push the stdout lines into the buffer, the stderr lines are not available...need to rework this a bit to show stderr lines when timing out as well.

joyeecheung:
Travis is now green (see https://travis-ci.org/nodejs/llnode/builds/308579054), although sometimes it fails because of https://github.com/nodejs/llnode/issues/150 .

This is now ready for review, cc @cjihrig @indutny @bnoordhuis @hhellyer , thanks!

test/common.js
```diff
@@ -72,109 +84,171 @@ SessionOutput.prototype._unqueueWait = function _unqueueWait() {
...
 
-    if (!regexp.test(line))
+  let done = false;
+  const check = setTimeout(() => {
+    if (done)
       return;
 
     self.removeListener('line', onLine);
     self._unqueueWait();
+    console.error(`${'='.repeat(10)} lldb output ${'='.repeat(10)}`);
```
cjihrig: Are these `console.error()`s intentional?

test/inspect-test.js
```diff
@@ -333,17 +351,21 @@ tape('v8 inspect', (t) => {
     }
   });
 
-  sess.linesUntil(/}>/, (lines) => {
-    lines = lines.join('\n');
-    t.ok(/internal fields/.test(lines), 'method.scopedAPI.internalFields');
-  });
+  if (process.version >= 'v5.0.0') {
```
cjihrig: Is this safe to do?

```
> process.version
'v10.0.0-pre'
> process.version >= 'v5.0.0'
false
```

test/usage-test.js
```diff
@@ -2,42 +2,42 @@
...
     sess.send('v8 print');
   });
 
-  sess.stderr.linesUntil(/USAGE/, (lines) => {
-    t.ok(/^error: USAGE: v8 print expr$/.test(removeBlankLines(lines)[0]),
-         'print usage message');
+  sess.stderr.linesUntil(/USAGE/, (err, lines) => {
+    t.error(err);
+    const re = /^error: USAGE: v8 print expr$/;
+    t.ok(lines.some(line => re.test(line.trim())), 'print usage message');
```
cjihrig: Since `lines.some(line => re.test(line.trim()))` is repeated a number of times, it might be better to keep it as a helper function that you can pass `re` and `lines` to.

test/common.js
joyeecheung: @cjihrig Yes, the intent is to log the lldb output that `wait` fails to match the expected regexp against, even with TEST_LLNODE_DEBUG turned off. After this log the error will be passed to the callback and the test should fail immediately.

test/inspect-test.js
joyeecheung: @cjihrig Oops, Node core jenkins scripts bug dejavu..this pattern is used in other places as well, I would like to fix them in a separate PR before v10.x releases.

joyeecheung:
@cjihrig Added a `containsLine` function to DRY the usage-test. The `process.version` comparison can be fixed in a later PR.

If there are no more reviews in 2 days I would like to land this so we can get a green Travis CI. The jenkins one probably need a bit of configuration update (last time I tried it it was using the old flow to build the plugin).

test/scan-test.js
```diff
@@ -1,56 +1,86 @@
...
-    sess.send(`target create -c ${common.core}`);
+    test(process.execPath, common.core, t);
   });
-
-  sess.wait(/Core file[^\n]+was loaded/, () => {
+}
+
+function test(executable, core, t) {
+  let sess, ranges;
+  if (process.env.LLNODE_NO_RANGES) {
```
joyeecheung: This is an environment variable for testing linux coredumps on Mac with newer versions of lldb (which does not need ranges file and setting one would result in empty scans). We can also just port the lldb version detection part from `configure` over and skip the ranges generation on higher versions of lldb (also that way we don't even to configure `TEST_LLNODE_BINARY` in the CI/on Linux anymore)

Makefile
```diff
@@ -19,9 +19,14 @@ uninstall-linux:
...
-	TEST_LLDB_BINARY=`which lldb-3.6` npm test
+configure: scripts/configure.js
+	node scripts/configure.js
+
+plugin: configure
+	./gyp_llnode
+	$(MAKE) -C out/
+
+_travis: plugin
+	TEST_LLDB_BINARY=`which lldb-3.9` TEST_LLNODE_DEBUG=true npm test
```
bnoordhuis: These targets are effectively .PHONY and will always run.  Is that okay?

joyeecheung: @bnoordhuis I think it's OK to make them .PHONY for now, we can write some proper make rules later.

PR #148, opened 2017-11-16
src: add total row to findjsobjects command

mmarchini:
Add a row with total objects count and total size to findjsobjects.

Demo: 
[asciicast](https://asciinema.org/a/147759)

### DIFF
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -62,6 +62,7 @@ bool FindObjectsCmd::DoExecute(SBDebugger d, char** cmd,
             TypeRecord::CompareInstanceCounts);
 
   uint64_t total_objects = 0;
+  uint64_t total_size = 0;
 
   result.Printf(" Instances  Total Size Name\n");
   result.Printf(" ---------- ---------- ----\n");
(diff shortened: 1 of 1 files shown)


joyeecheung:
BTW: the CI is half-broken right now, needs https://github.com/nodejs/llnode/pull/144 and https://github.com/nodejs/github-bot/pull/158 to fix that, so ignore that yellow "Some checks haven’t completed yet".

src/llscan.cc
```diff
@@ -61,7 +61,7 @@ bool FindObjectsCmd::DoExecute(SBDebugger d, char** cmd,
   std::sort(sorted_by_count.begin(), sorted_by_count.end(),
             TypeRecord::CompareInstanceCounts);
 
-  uint64_t total_objects = 0;
+  uint64_t total_objects = 0, total_size = 0;
```
indutny: Minor nit: do you mind splitting this into two separate declarations?

mmarchini: No problem. PR updated with this change

indutny: Thank you!

Issue #149, opened 2017-11-16
Detailed findjsobjects output


mmarchini:
I was looking at the mdb_v8 output, and they have quite a different output for this command. For example, instead of grouping objects only by type or by constructor name,  they also use the object properties and array indexes to aggregate them. They also have three other columns (Representative Object, Number of Properties and list of Properties) in their output, which looks like this:

```
OBJECT #OBJECTS   #PROPS CONSTRUCTOR: PROPS
   │       │         │        │         │
   │       │         │        │         └───> The first few properties for those
   │       │         │        │               objects
   │       │         │        │
   │       │         │        └─────────────> Name of this type's constructor
   │       │         │
   │       │         └──────────────────────> Number of properties + array 
   │       │                                  indexes for those objects
   │       │                                  
   │       └────────────────────────────────> Number of instances
   │       
   └────────────────────────────────────────> Representative object (address to 
                                              one sample object from this group)
   
   
```

I developed a prototype for a detailed version of findjsobjects, which is available [here (branch feature/findjsobjects-detailed-output)](https://github.com/mmarchini/llnode) and can be used with `v8 findjsobjects --detailed`. There's also a [demo](https://asciinema.org/a/147794) on Asciinema.

In this prototype, I added three new columns (Representative Object, Number of Properties and Array Size), and added the first three properties to the objects group name (last column). Objects are grouped by constructor name + properties + array size.

If you're interested I can create a PR with those changes. 
I'm also opened to suggestions on this.

bnoordhuis:
Nice!

I can think of a few improvements (print length rather than "length" for arrays, make symbol properties display nicely, use an object's `@@toStringTag` if there is one, etc.) but this is a good start.

mmarchini:
Thanks for the feedback!

I'll take a look into that this week

mmarchini:
> print length rather than "length" for arrays

Did you mean using `JSObject::GetArrayLength()` instead of `JSArray::Length`?

> make symbol properties display nicely, use an object's @@toStringTag if there is one

I did some digging into that but seems like llnode doesn't support Symbol properties or `@@toStringTag` yet (I'll open another issue to investigate it in the future). I like the idea of using `@@toStringTag` when it's present, but what should we do when `@@toStringTag` is a function? For example:

```javascript
class Foo {
  constructor(yes) {
    this.bar = "foo bar";
  }
  get [Symbol.toStringTag]() {
    return "[ Foo " + this.bar + " ]";
  }
}
```

[...]

Issue #150, opened 2017-11-28
test: the uint8-array in inspect-scenario sometimes gets neutered


joyeecheung:
```
ok 13 addr of `Object` should match
[TEST] [LINE]     [0]=0x0000134ccad82201:<null>,
[TEST] [LINE]     [4]=0x0000134ccad822d1:<undefined>,
[TEST] [LINE]     [23]=0x0000134c51ddf5d9:<JSRegExp source=/regexp/>,
[TEST] [LINE]     [25]=0x0000134cce90b1c9:<function: c.hashmap.(anonymous function) at /Users/joyee/projects/llnode/test/fixtures/inspect-scenario.js:63:19>}
[TEST] [LINE]   properties {
[TEST] [LINE]     .some-key=<Smi: 42>,
[TEST] [LINE]     .other-key=0x0000134cce907fd1:<String: "ohai">,
[TEST] [LINE]     .cons-string=0x0000134c51dde779:<String: "this could be a ...">,
[TEST] [LINE]     .internalized-string=0x0000134cce908099:<String: "foobar">,
[TEST] [LINE]     .thin-string=0x0000134c51dde821:<String: "foobar">,
[TEST] [LINE]     .externalized-string=0x0000134cce908151:<String: "(external)">,
[TEST] [LINE]     .sliced-externalized-string=0x0000134c51dde9e9:<String: "(external)">,
[TEST] [LINE]     .array=0x0000134c51ddeb19:<Array: length=6>,
[TEST] [LINE]     .long-array=0x0000134c51ddeca9:<Array: length=20>,
[TEST] [LINE]     .array-buffer=0x0000134c51ddeff1:<ArrayBuffer: backingStore=0x0000000102700ef0, byteLength=5>,
[TEST] [LINE]     .uint8-array=0x0000134c51ddf0b9:<ArrayBufferView [neutered]>,
[TEST] [LINE]     .buffer=0x0000134c51ddf431:<ArrayBufferView: backingStore=0x000000010280d600, byteOffset=2048, byteLength=6>,
[TEST] [LINE]     .scoped=<unknown field type>}>
ok 14 [0] null element
ok 15 [4] undefined element
ok 16 [23] RegExp element
ok 17 [25] Arrow Function element
ok 18 .some-key property
ok 19 .other-key property
ok 20 .array JSArray property
ok 21 .array JSArray property
ok 22 .array-buffer JSArrayBuffer property
not ok 23 .uint8-array JSArrayBufferView property
  ---
[...]

PR #155, opened 2017-12-08
src: add a detailed output to findjsobjects

mmarchini:
Add a detailed output to findjsobjects. Inspired on mdb_v8 output, which
prints a representative object, the number of properties and array
indexes, and the first few properties for a map. This option will also
group types by its properties instead of only grouping by constructor
name.

Fixes: https://github.com/nodejs/llnode/issues/149

### DIFF
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -1192,14 +1192,16 @@ void FindJSObjectsVisitor::InsertOnDetailedMapsToInstances(
 
   auto type_name_with_properties = map_info.GetTypeNameWithProperties();
 
-  auto entry = std::make_pair(*type_name_with_properties, nullptr);
+  auto entry = std::make_pair(type_name_with_properties, nullptr);
   auto pp = &llscan_->GetDetailedMapsToInstances().insert(entry).first->second;
   // No entry in the map, create a new one.
-  if (*pp == nullptr)
-    *pp = new DetailedTypeRecord(*map_info.GetTypeNameWithProperties(
-                                     MapCacheEntry::kDontShowArrayLength, 3),
+  if (*pp == nullptr) {
+    auto type_name_with_three_properties = map_info.GetTypeNameWithProperties(
+        MapCacheEntry::kDontShowArrayLength, 3);
+    *pp = new DetailedTypeRecord(type_name_with_three_properties,
                                  map_info.own_descriptors_count_,
                                  map_info.indexed_properties_count_);
+  }
   t = *pp;
 
   // Determine if this is a new instance.
@@ -1280,26 +1282,24 @@ bool LLScan::ScanHeapForObjects(lldb::SBTarget target,
   return true;
 }
 
-std::unique_ptr<std::string>
+std::string
 FindJSObjectsVisitor::MapCacheEntry::GetTypeNameWithProperties(
     ShowArrayLength show_array_length, size_t max_properties) {
-  std::unique_ptr<std::string> type_name_with_properties(
-      new std::string(type_name));
+  std::string type_name_with_properties(type_name);
 
   if (show_array_length == kShowArrayLength) {
-    type_name_with_properties->append(
-        "[" + std::to_string(indexed_properties_count_) + "]");
+    type_name_with_properties +=
+        "[" + std::to_string(indexed_properties_count_) + "]";
   }
 
   size_t i = 0;
   max_properties = max_properties ? std::min(max_properties, properties_.size())
                                   : properties_.size();
   for (auto it = properties_.begin(); i < max_properties; ++it, i++) {
-    std::string property = *it;
-    type_name_with_properties->append((i ? ", " : ": ") + property);
+    type_name_with_properties += (i ? ", " : ": ") + *it;
   }
   if (max_properties < properties_.size()) {
-    type_name_with_properties->append(", ...");
+    type_name_with_properties += ", ...";
   }
 
   return type_name_with_properties;
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -1197,7 +1197,8 @@ void FindJSObjectsVisitor::InsertOnDetailedMapsToInstances(
   // No entry in the map, create a new one.
   if (*pp == nullptr) {
     auto type_name_with_three_properties = map_info.GetTypeNameWithProperties(
-        MapCacheEntry::kDontShowArrayLength, 3);
+        MapCacheEntry::kDontShowArrayLength,
+        kNumberOfPropertiesForDetailedOutput);
     *pp = new DetailedTypeRecord(type_name_with_three_properties,
                                  map_info.own_descriptors_count_,
                                  map_info.indexed_properties_count_);
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -48,6 +48,22 @@ bool FindObjectsCmd::DoExecute(SBDebugger d, char** cmd,
     return false;
   }
 
+  v8::Value::InspectOptions inspect_options;
+  inspect_options.detailed;
+  ParseInspectOptions(cmd, &inspect_options);
+
+  if (inspect_options.detailed) {
+    DetailedOutput(result);
+  } else {
+    SimpleOutput(result);
+  }
+
+  result.SetStatus(eReturnStatusSuccessFinishResult);
+  return true;
+}
+
+
+void FindObjectsCmd::SimpleOutput(SBCommandReturnObject& result) {
   /* Create a vector to hold the entries sorted by instance count
    * TODO(hhellyer) - Make sort type an option (by count, size or name)
    */
@@ -78,9 +94,39 @@ bool FindObjectsCmd::DoExecute(SBDebugger d, char** cmd,
 
   result.Printf(" ---------- ---------- \n");
   result.Printf(" %10" PRId64 " %10" PRId64 " \n", total_objects, total_size);
+}
 
-  result.SetStatus(eReturnStatusSuccessFinishResult);
-  return true;
+
+void FindObjectsCmd::DetailedOutput(SBCommandReturnObject& result) {
+  std::vector<DetailedTypeRecord*> sorted_by_count;
+  for (auto it : llscan.GetDetailedMapsToInstances()) {
+    sorted_by_count.push_back(it.second);
+  }
+
+  std::sort(sorted_by_count.begin(), sorted_by_count.end(),
+            TypeRecord::CompareInstanceCounts);
+  uint64_t total_objects = 0;
+  uint64_t total_size = 0;
+
+  result.Printf(
+      "   Repr. Obj.  Instances  Total Size  Properties  Array Size Name\n");
+  result.Printf(
+      " ------------ ---------- ----------- ----------- ----------- ----\n");
+
+  for (auto t : sorted_by_count) {
+    result.Printf(" %12" PRIx64 " %10" PRId64 " %11" PRId64 " %11" PRId64
+                  " %11" PRId64 " %s\n",
+                  *(t->GetInstances().begin()), t->GetInstanceCount(),
+                  t->GetTotalInstanceSize(), t->GetOwnDescriptorsCount(),
+                  t->GetIndexedPropertiesCount(), t->GetTypeName().c_str());
+    total_objects += t->GetInstanceCount();
+    total_size += t->GetTotalInstanceSize();
+  }
+
+  result.Printf(
+      " ------------ ---------- ----------- ----------- ----------- ----\n");
+  result.Printf("             %11" PRId64 " %11" PRId64 " \n", total_objects,
+                total_size);
 }
 
 
@@ -1062,9 +1108,8 @@ ReferencesVector* FindReferencesCmd::StringScanner::GetReferences() {
 }
 
 
-FindJSObjectsVisitor::FindJSObjectsVisitor(SBTarget& target,
-                                           TypeRecordMap& mapstoinstances)
-    : target_(target), mapstoinstances_(mapstoinstances) {
+FindJSObjectsVisitor::FindJSObjectsVisitor(SBTarget& target, LLScan* llscan)
+    : target_(target), llscan_(llscan) {
   found_count_ = 0;
   address_byte_size_ = target_.GetProcess().GetAddressByteSize();
   // Load V8 constants from postmortem data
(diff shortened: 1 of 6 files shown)


src/llscan.cc
```diff
@@ -1141,6 +1163,56 @@ uint64_t FindJSObjectsVisitor::Visit(uint64_t location, uint64_t word) {
...
+  TypeRecord* t;
+
+  /* No entry in the map, create a new one. */
+  if (llscan_->GetMapsToInstances().count(map_info.type_name) == 0) {
+    t = new TypeRecord(map_info.type_name);
+    llscan_->GetMapsToInstances().emplace(map_info.type_name, t);
+  } else {
+    /* Get entry from the map. */
+    t = llscan_->GetMapsToInstances().at(map_info.type_name);
+  }
```
bnoordhuis: I realize this comes from existing code but `.count()` is an O(n) function.  `.find()` would be better but since you're inserting, this is even more concise and efficient:
```c++
auto entry = std::make_pair(map_info.type_name, nullptr);
auto pp = &llscan_->GetMapsToInstances().insert(entry).first->second;
if (*pp == nullptr) *pp = new TypeRecord(map_info.type_name);
t = *pp;
```

src/llscan.h
```diff
@@ -174,36 +181,71 @@ class TypeRecord {
...
+        indexed_properties_count_(indexed_properties_count) {}
+  inline uint64_t GetOwnDescriptorsCount() { return own_descriptors_count_; };
+  inline uint64_t GetIndexedPropertiesCount() {
+    return indexed_properties_count_;
+  };
+
+ private:
+  std::vector<std::string> properties_;
+  uint64_t own_descriptors_count_ = 0;
+  uint64_t indexed_properties_count_ = 0;
```
bnoordhuis: The initializers are pointless, aren't they, since they're assigned in the constructor's init list?

src/llscan.h
```diff
@@ -174,36 +181,71 @@ class TypeRecord {
...
 
+class DetailedTypeRecord : public TypeRecord {
+ public:
+  DetailedTypeRecord(std::string& type_name, uint64_t own_descriptors_count,
+                     uint64_t indexed_properties_count)
+      : TypeRecord(type_name),
+        own_descriptors_count_(own_descriptors_count),
+        indexed_properties_count_(indexed_properties_count) {}
+  inline uint64_t GetOwnDescriptorsCount() { return own_descriptors_count_; };
+  inline uint64_t GetIndexedPropertiesCount() {
```
bnoordhuis: These two methods could be `const` and you don't have to mark them `inline` since they have an inline definition.

src/llscan.h
```diff
@@ -174,36 +181,71 @@ class TypeRecord {
...
 
 class FindJSObjectsVisitor : MemoryVisitor {
  public:
-  FindJSObjectsVisitor(lldb::SBTarget& target, TypeRecordMap& mapstoinstances);
+  struct MapCacheEntry {
+    std::string type_name;
+    std::vector<std::string> properties;
+    uint64_t own_descriptors_count = 0;
+    uint64_t indexed_properties_count = 0;
+    bool is_histogram;
```
bnoordhuis: Can you give these an underscore suffix?

src/llscan.h
```diff
@@ -174,36 +181,71 @@ class TypeRecord {
...
+
+    std::unique_ptr<std::string> GetTypeNameWithProperties(
+        bool show_array_length = true, size_t max_properties = 0);
```
bnoordhuis: Consider using an `enum ShowArrayLength { SHOW_ARRAY_LENGTH, DONT_SHOW_ARRAY_LENGTH }` for this, that's more obvious than a bool at call sites.

src/llscan.cc
```diff
@@ -1141,6 +1163,56 @@ uint64_t FindJSObjectsVisitor::Visit(uint64_t location, uint64_t word) {
...
+  /* No entry in the map, create a new one. */
+  if (llscan_->GetDetailedMapsToInstances().count(*type_name_with_properties) ==
+      0) {
+    t = new DetailedTypeRecord(*map_info.GetTypeNameWithProperties(false, 3),
+                               map_info.own_descriptors_count,
+                               map_info.indexed_properties_count);
+    llscan_->GetDetailedMapsToInstances().emplace(*type_name_with_properties,
+                                                  t);
+  } else {
+    /* Get entry from the map. */
```
bnoordhuis: Feel free to drop these comments, it's almost at the level of `i += 1;   // add 1 to i`

src/llscan.h
mmarchini: Yes, they are now. They were needed in a previous version and I forgot to remove them.

src/llscan.h
mmarchini: Sure. Can I give it to `type_name` and `is_histogram` too? They were already there before.

src/llscan.h
```diff
@@ -8,19 +8,27 @@
...
+class LLScan;
+
 typedef std::vector<uint64_t> ReferencesVector;
 
 typedef std::map<uint64_t, ReferencesVector*> ReferencesByValueMap;
 typedef std::map<std::string, ReferencesVector*> ReferencesByPropertyMap;
 typedef std::map<std::string, ReferencesVector*> ReferencesByStringMap;
 
 
+enum ShowArrayLength { SHOW_ARRAY_LENGTH, DONT_SHOW_ARRAY_LENGTH };
```
mmarchini: Questions:
* Should this be inside `FindJsObjectVisitor::MapCacheEntry` or it's ok to keep it global?
* There's another `enum` in this file, and the enumerators' names use this format: `kCamelCasedName`. Should I use the same format (kShowArrayLength, kDontShowArrayLength)?

joyeecheung: @mmarchini I would suggest to put it inside `FindJsObjectVisitor::MapCacheEntry` or `FindJsObjectVisitor`, and yes it should be named in the camel case style.

mmarchini:
`enum` names updated and test case added.

mmarchini:
CI error is unrelated (#150)

mmarchini:
Ping @joyeecheung @bnoordhuis

joyeecheung:
@mmarchini ACK, will try to look into it during the weekend

src/llscan.cc
```diff
@@ -1202,14 +1270,77 @@ bool LLScan::ScanHeapForObjects(lldb::SBTarget target,
...
+    FindJSObjectsVisitor v(target, this);
 
     ScanMemoryRanges(v);
   }
 
   return true;
 }
 
+std::unique_ptr<std::string>
+FindJSObjectsVisitor::MapCacheEntry::GetTypeNameWithProperties(
```
bnoordhuis: You don't have to wrap it in a unique_ptr, the compiler will elide the copy when you return the std::string by value.

(Compilers are not required to by the spec but they all do it.)

src/llscan.cc
```diff
@@ -1202,14 +1270,77 @@ bool LLScan::ScanHeapForObjects(lldb::SBTarget target,
...
+    type_name_with_properties->append(
+        "[" + std::to_string(indexed_properties_count_) + "]");
+  }
+
+  int i = 0;
+  max_properties = max_properties ? std::min(max_properties, properties_.size())
+                                  : properties_.size();
+  for (auto it = properties_.begin(); i < max_properties; ++it, i++) {
+    std::string property = *it;
+    type_name_with_properties->append((i ? ", " : ": ") + property);
```
bnoordhuis: I'd use two append statements here to avoid creating a temporary string.

src/llscan.cc
```diff
@@ -1202,14 +1270,77 @@ bool LLScan::ScanHeapForObjects(lldb::SBTarget target,
...
+                                               v8::Error& err) {
+  // Check type first
+  is_histogram = FindJSObjectsVisitor::IsAHistogramType(map, err);
+
+  // On success load type name
+  if (is_histogram) type_name = heap_object.GetTypeName(err);
+
+  v8::HeapObject descriptors_obj = map.InstanceDescriptors(err);
+  if (err.Fail()) {
+    return false;
```
bnoordhuis: On a single line for consistency with the other guards in this function?

src/llscan.h
```diff
@@ -174,36 +181,75 @@ class TypeRecord {
 
 
  private:
+  friend DetailedTypeRecord;
```
bnoordhuis: `friend class`

src/llscan.h
```diff
@@ -174,36 +181,75 @@ class TypeRecord {
...
 };
 
+class DetailedTypeRecord : public TypeRecord {
+ public:
+  DetailedTypeRecord(std::string& type_name, uint64_t own_descriptors_count,
+                     uint64_t indexed_properties_count)
+      : TypeRecord(type_name),
+        own_descriptors_count_(own_descriptors_count),
+        indexed_properties_count_(indexed_properties_count) {}
+  const uint64_t GetOwnDescriptorsCount() { return own_descriptors_count_; };
```
bnoordhuis: `uint6_t own_descriptors_count() const { ... }` since it's a simple getter and note that the `const` should come after the method name.  Putting it on the type is a no-op and one that some compilers complain about.

edit: oh, I see there are some existing baroque getter names.  Not wrong then.

src/llscan.h
```diff
@@ -174,36 +181,75 @@ class TypeRecord {
...
+  uint64_t indexed_properties_count_;
+};
+
 typedef std::map<std::string, TypeRecord*> TypeRecordMap;
+typedef std::map<std::string, DetailedTypeRecord*> DetailedTypeRecordMap;
 
 class FindJSObjectsVisitor : MemoryVisitor {
  public:
-  FindJSObjectsVisitor(lldb::SBTarget& target, TypeRecordMap& mapstoinstances);
+  struct MapCacheEntry {
```
bnoordhuis: `MapCacheEntry` can still be private, can't it?

src/llscan.h
```diff
@@ -174,36 +181,75 @@ class TypeRecord {
...
   uint64_t Visit(uint64_t location, uint64_t word);
 
   uint32_t FoundCount() { return found_count_; }
 
- private:
-  struct MapCacheEntry {
-    std::string type_name;
-    bool is_histogram;
-  };
+  static bool IsAHistogramType(v8::Map& map, v8::Error& err);
```
bnoordhuis: Same for `IsAHistogramType()`.

src/llscan.h
```diff
@@ -174,36 +181,75 @@ class TypeRecord {
...
+  void InsertOnDetailedMapsToInstances(
+      uint64_t word, v8::Map map, FindJSObjectsVisitor::MapCacheEntry map_info,
+      v8::Error& err);
 
   lldb::SBTarget& target_;
   uint32_t address_byte_size_;
   uint32_t found_count_;
 
-  TypeRecordMap& mapstoinstances_;
+  LLScan* llscan_;
```
bnoordhuis: Maybe `LLSCan* const llscan_;`?

src/llscan.cc
```diff
@@ -78,9 +94,39 @@ bool FindObjectsCmd::DoExecute(SBDebugger d, char** cmd,
...
   result.Printf(" ---------- ---------- \n");
   result.Printf(" %10" PRId64 " %10" PRId64 " \n", total_objects, total_size);
+}
 
-  result.SetStatus(eReturnStatusSuccessFinishResult);
-  return true;
+
+void FindObjectsCmd::DetailedOutput(SBCommandReturnObject& result) {
+  std::vector<DetailedTypeRecord*> sorted_by_count;
+  for (auto it : llscan.GetDetailedMapsToInstances()) {
```
joyeecheung: `s/it/kv/`? It's not an iterator.

src/llscan.cc
```diff
@@ -78,9 +94,39 @@ bool FindObjectsCmd::DoExecute(SBDebugger d, char** cmd,
...
+    sorted_by_count.push_back(it.second);
+  }
+
+  std::sort(sorted_by_count.begin(), sorted_by_count.end(),
+            TypeRecord::CompareInstanceCounts);
+  uint64_t total_objects = 0;
+  uint64_t total_size = 0;
+
+  result.Printf(
+      "   Repr. Obj.  Instances  Total Size  Properties  Array Size Name\n");
```
joyeecheung: The name `Repr. Obj.` does not seem very intuitive here. Maybe `Sample`? Also the name `Elements` seems to be more accurate than `Array Size` since we are talking about the number of properties & elements here.

src/llscan.cc
```diff
@@ -1092,13 +1137,7 @@ uint64_t FindJSObjectsVisitor::Visit(uint64_t location, uint64_t word) {
...
   MapCacheEntry map_info;
   if (map_cache_.count(map.raw()) == 0) {
-    // Check type first
-    map_info.is_histogram = IsAHistogramType(map, err);
-
-    // On success load type name
-    if (map_info.is_histogram)
-      map_info.type_name = heap_object.GetTypeName(err);
-
+    if (!map_info.Load(map, heap_object, err)) return address_byte_size_;
```
joyeecheung: The prevalent style here would be:

```c++
map_info.Load(map, heap_object, err);
if (err.Fail()) {
  return address_byte_size_;
}
```

src/llscan.cc
```diff
@@ -1141,6 +1164,51 @@ uint64_t FindJSObjectsVisitor::Visit(uint64_t location, uint64_t word) {
...
+  DetailedTypeRecord* t;
+
+  auto type_name_with_properties = map_info.GetTypeNameWithProperties();
+
+  auto entry = std::make_pair(*type_name_with_properties, nullptr);
+  auto pp = &llscan_->GetDetailedMapsToInstances().insert(entry).first->second;
+  // No entry in the map, create a new one.
+  if (*pp == nullptr)
+    *pp = new DetailedTypeRecord(*map_info.GetTypeNameWithProperties(
+                                     MapCacheEntry::kDontShowArrayLength, 3),
```
joyeecheung: It might worth making `3` a static variable in `FindJSObjectsVisitor` (ideally it can be an option)

mmarchini:
Updated :+1:

joyeecheung:
CI is green, will merge this later today.

joyeecheung:
Landed in 20b14d967ecfe60a3b778611de04bad4f59b14b9, thanks!

PR #162, opened 2018-01-02
docs: update installation, loading, developing and testing instructions

joyeecheung:
- doc: add npm badge and use flat square style
- doc: update installation and loading instructions
  - Update the prerequisites for each OS, on macOS one can use
    brew to install custom versions of lldb, also the shared library
    should be installed as well.
  - It's no longer necessary to download GYP and lldb headers manually
    since the configure scripts would handle all of that
  - Mention using `lldbinit` to load the plugin automatically, list
    different approaches to load the plugin.
- doc: add section about developing and testing

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,142 +3,118 @@
 [![npm](https://img.shields.io/npm/v/llnode.svg?style=flat-square)](https://npmjs.org/package/llnode)
 [![Build Status](https://img.shields.io/travis/nodejs/llnode.svg?style=flat-square)](https://travis-ci.org/nodejs/llnode)
 
-Node.js v4.x-v8.x C++ plugin for the [LLDB](http://lldb.llvm.org) debugger.
+Node.js v4.x+ C++ plugin for the [LLDB](http://lldb.llvm.org) debugger.
 
 The llnode plugin adds the ability to inspect JavaScript stack frames, objects,
 source code and more to the standard C/C++ debugging facilities when working
 with Node.js processes or core dumps in LLDB.
 
+## Demo
+
+https://asciinema.org/a/29589
+
 ### Quick start
 
-Start an lldb session with the llnode plugin pre-loaded:
+Start an LLDB session with the llnode plugin automatically loaded:
 
 ```bash
 npm install -g llnode
-llnode node -c core
+llnode `which node` -c /path/to/core/dump
 ```
 
 - For more details on starting llnode see the [Usage](#usage) section.
 - To get started with the llnode commands see the [Commands](#commands) section.
 
-## Demo
-
-https://asciinema.org/a/29589
-
 ## Install Instructions
 
-To use llnode you need to have the LLDB debugger installed.
+### Prerequisites: Install LLDB and its Library
 
-- On OS X lldb is installed as part of Xcode. You will need Xcode both to build and run llnode.
-- On Linux install the lldb package using your distributions package manager.
+To use llnode you need to have the LLDB debugger installed. The recommended
+version is LLDB 3.9 and above.
 
-### Global install with npm
+- OS X/macOS
+  - You can install [Xcode](https://developer.apple.com/xcode/) and use the
+    LLDB that comes with it.
+  - Optionally, you can install newer versions of lldb using Homebrew with
+
+    ```bash
+    brew update && brew install --with-lldb --with-toolchain llvm
+    ```
+    
+    and make sure `/usr/local/opt/llvm/bin` gets searched before `/usr/bin/`
+    on your `PATH`. llnode will link to the LLDB installation returned by
+    `llvm-config` if available.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,183 +1,139 @@
...
-`make install-osx` build step above, or if installing
-with npm copy `node_modules/llnode/llnode.dylib` to
-` ~/Library/Application\ Support/LLDB/PlugIns/`.
+Similar to the `~/.lldbinit` approach, this way LLDB will also load the plugin
+automatically on start-up. Doing this may require additional permissions
+to be able to copy the plugin libary to the system plugin directory.
 
-### Linux
+On Linux, run `make install-linux` in the project directory, or if
+installing with npm, copy `node_modules/llnode/llnode.so`
```
mmarchini: Maybe we can suggest linking instead of (or in addition to) copying?

README.md
```diff
@@ -259,6 +215,85 @@ The following subcommands are supported:
...
+
+```bash
+TEST_LLDB_BINARY=`which lldb-4.0` npm run test
+```
+
+### Useful Environment Variables
+
+* `LLNODE_DEBUG=true` to see additional debug info from llnode 
+* `TEST_LLNODE_DEBUG=true` to see additional debug info coming from the tests
+* `LLNODE_CORE=<path-to-core-dump> LLNODE_NODE_EXE=<path-to-node> LLNODE_NO_RANGES=true`
```
mmarchini: `s/path-to-core-dump/path-to-core-dump-from-test-scenario/` to make it clearer?

README.md
```diff
@@ -1,183 +1,139 @@
 # llnode
 
-[![Build Status](https://travis-ci.org/nodejs/llnode.svg?branch=master)](https://travis-ci.org/nodejs/llnode)
+[![npm](https://img.shields.io/npm/v/node-core-utils.svg?style=flat-square)](https://npmjs.org/package/node-core-utils)
```
cjihrig: This links to `node-core-utils`

README.md
```diff
@@ -1,183 +1,139 @@
...
+[![Build Status](https://img.shields.io/travis/nodejs/node-core-utils.svg?style=flat-square)](https://travis-ci.org/nodejs/node-core-utils)
```
cjihrig: Same here

README.md
```diff
@@ -1,183 +1,139 @@
...
-gmake -C out/ -j9
-```
-
-(The LLDB function ComputeSystemPluginsDirectory is not implemented on FreeBSD.
-The plugin library must be loaded manually.)
+If you install llnode globally via npm (`npm install -g llnode`), you can use
+the `llnode` shortcut script. This starts `lldb` and automatically issues
+the `plugin load` command. All parameters to the llnode script are passed
+directly to lldb. If you do not do a local install the shortcut will be in
+`node_modules/.bin/llnode`
```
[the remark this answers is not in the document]
joyeecheung: I just tried `npx llnode` with v8.9.3 and it does not seem to work properly on macOS...I'd rather leave it there.

README.md
joyeecheung: @cjihrig Ooops, copy-paste error...fixed, thanks!

README.md
```diff
@@ -1,183 +1,139 @@
...
 
-### Global install with npm
+To use llnode you need to have the LLDB debugger installed. The recommended
+version is LLDB 3.9 and above.
 
-```bash
-npm install -g llnode
-```
+- OS X/macOS
+  - You can install LLDB as part of [Xcode](https://developer.apple.com/xcode/).
```
bnoordhuis: Maybe `use the LLDB that is part of` instead of `install LLDB as part of`?

The "you can" suggests it's optional or requires extra steps but I don't recall an option to disable it when I last installed xcode.

README.md
```diff
@@ -259,6 +215,85 @@ The following subcommands are supported:
...
+
+To debug `test/scan-test.js` with a prepared core dump:
+
+```
+LLNODE_DEBUG=true TEST_LLNODE_DEBUG=true \
+  LLNODE_CORE=/path/to/core/dump/of/inspect/scenario.js \
+  LLNODE_NODE_EXE=/path/to/node \
+  LLNODE_NO_RANGES=true \
+  node test/scan-test.js
+```
```
bnoordhuis: ^ Good addition, this section.

README.md
joyeecheung: @bnoordhuis I will update it to `You can install [Xcode](https://developer.apple.com/xcode/) and use the LLDB that comes with it` so the action that needs to be done is clearer.

joyeecheung:
@bnoordhuis @cjihrig @mmarchini Thanks for the reviews, I plan to merge this later today and do a new release with the new documentation.

PR #167, opened 2018-02-04
test: move core saving and loading into common.js

joyeecheung:
This was used by https://github.com/nodejs/llnode/pull/147 to reuse the code that generates a core dump and saves it for testing, but it's worth getting merged first.

cc @cjihrig @bnoordhuis

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -224,11 +224,11 @@ exports.saveCore = function saveCore(options, cb) {
     }
     sess.quit();
 
-    if (!process.env.LLNODE_NO_RANGES) {
-      const ranges = core + '.ranges';
-      exports.generateRanges(core, ranges, cb);
-    } else {
+    if (process.env.LLNODE_NO_RANGES) {
       cb(null);
+    } else {
+      const ranges = core + '.ranges';
+      exports.generateRanges(core, ranges, cb); 
     }
   });
 }
@@ -267,11 +267,6 @@ Session.prototype.quit = function quit() {
 
   this.send('quit');
 };
-
-Session.prototype.quitWithCore = function quitWithCore() {
-  this.send('target delete 0');
-  this.quit();
-}
 
 Session.prototype.send = function send(line, callback) {
   debug(`[SEND][${this.lldb.pid}]`, line);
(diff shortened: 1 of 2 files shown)


test/common.js
```diff
@@ -204,12 +204,49 @@ Session.create = function create(scenario) {
...
 };
 
-Session.loadCore = function loadCore(executable, core, ranges) {
-  return new Session({
+exports.saveCore = function saveCore(options, cb) {
+  const scenario = options.scenario;
+  const core = options.core || exports.core;
+
+  // Create a core and test
+  let sess = Session.create(scenario);
```
cjihrig: Can this be `const`?

test/common.js
```diff
@@ -204,12 +204,49 @@ Session.create = function create(scenario) {
...
+  let ranges;
```
cjihrig: Is this used anywhere? I see `const ranges = ...` later on.

joyeecheung:
@cjihrig Thanks for the review. Updated, PTAL.

test/common.js
```diff
@@ -204,12 +204,48 @@ Session.create = function create(scenario) {
...
+      return cb(err);
+    }
+    sess.quit();
+
+    if (!process.env.LLNODE_NO_RANGES) {
+      const ranges = core + '.ranges';
+      exports.generateRanges(core, ranges, cb);
+    } else {
+      cb(null);
+    }
```
bnoordhuis: Can you swap the branches?  No negation is never less easy to parse for humans, if you catch my drift.

test/common.js
```diff
@@ -232,8 +268,13 @@ Session.prototype.quit = function quit() {
   this.send('quit');
 };
 
+Session.prototype.quitWithCore = function quitWithCore() {
+  this.send('target delete 0');
```
bnoordhuis: That doesn't create a core file, does it?  It just unloads the image from lldb.

joyeecheung:
@bnoordhuis I think `target delete` is no longer necessary now that we kill the process after saving the core...updated, PTAL, thanks!

PR #170, opened 2018-03-09
src: do not create global llscan or global llv8

joyeecheung:
### test: fix hanging when testing with prepared core

When the session is created by loading a prepared core dump,
it should delete the target before quitting otherwise the
process will prompt about deleting it and will hang if it
does not receive an answer.

### src: do not create global llscan or global llv8

- Eliminates all references to the global llscan in the commands,
  only instantiate one LLScan instance in llnode.cc and pass it
  down to all commands.
- Eliminates all references to the global llv8, store it inside
  the LLScan instead, because LLScan must need a valid llv8
  (with postmortem metadata loaded) to function.
- Only instantiate a static LLScan and a static LLV8 during plugin
  initialization.

The first commit is a test fix, the second one is spinning off from https://github.com/nodejs/llnode/pull/147 so llscan.o can be linked by another module.

### DIFF
--- a/src/llnode.cc
+++ b/src/llnode.cc
@@ -27,8 +27,6 @@ using lldb::SBValue;
 using lldb::eReturnStatusFailed;
 using lldb::eReturnStatusSuccessFinishResult;
 
-v8::LLV8 llv8;
-
 char** CommandBase::ParseInspectOptions(char** cmd,
                                         v8::Value::InspectOptions* options) {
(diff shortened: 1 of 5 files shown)


joyeecheung:
cc @bnoordhuis @cjihrig

joyeecheung:
Only a known flake on Travis with Node.js v9.x. Merging..

PR #177, opened 2018-03-18
Split README, add more instructions and a few fixes to make llnode easier to contribute

joyeecheung:
### configure: skip linking tools/gyp if it exists

Use the existing tools/gyp if it exists. It could be checked out
from source instead of being linked to the node-gyp installation.
Also corrects the names in the gyp linking routines.

Also tries to find the built library in out/Release/obj.target/
if it's not in out/Release/ (the original gyp does not move it).

### docs: split the README and update CONTRIBUTING.md

- Split the README into install.md, usage.md and develop.md under docs/.
- Added OS-specific instructions on how to set up the prerequisites
- Update the Contributing guide so it applies to llnode instead of
  node core.
- Added instruction on how to install and run clang-format. Prefer
 the one available in the LLVM website.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 This document will guide you through the contribution process.
 
-### Step 1: Fork
+## Step 1: Fork
 
 Fork the project [on GitHub](https://github.com/indutny/llnode) and check out
 your copy locally.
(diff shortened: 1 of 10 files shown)


joyeecheung:
BTW the clang-format in the latest depot_tools is `clang-format version 5.0.0 (trunk 302580)`

bnoordhuis:
Sorry, seems I overlooked the notification for this PR.  Is this subsumed by #182 and/or #185?

joyeecheung:
@bnoordhuis The first commit is not necessary after #182 and the third commit needs a bit of rewrite after #185. The second commit is kind of trivial. I will open another doc PR after I resolve #185.

PR #178, opened 2018-03-21
findrefs: refactor ::PrintRefs template strings

mmarchini:
I noticed that we have only a few template strings used in all PrintRefs, and they were repeating multiple times in the code.


bnoordhuis:
(whispers) `static const char *const` or `static const char[]`

mmarchini:
Done!

Oh, I got some unrelated changes from `make format`. Should I keep them?

joyeecheung:
@mmarchini Which version of clang-format are you using?

mmarchini:
```bash
$ clang-format --version
clang-format version 5.0.1 (tags/RELEASE_501/final)
```

Issue #181, opened 2018-03-29
Drop support for the original gyp?


joyeecheung:
I was trying to merge `gyp_llnode` into `scripts/configure.js` and make `llnode.gyp` just `binding.gyp` (also in preparation for https://github.com/nodejs/llnode/pull/147). I noticed that it would be easier if we just drop support for the original gyp and only supports node-gyp, which handles a lot of configuration done by `gyp_llnode`.

Note that on master, cloning`https://chromium.googlesource.com/external/gyp.git` to `tools/gyp` and use that to build llnode does not actually work at the moment, and it always gets deleted and relinked to the gyp in the node-gyp installation. (the first commit in https://github.com/nodejs/llnode/pull/177 can fix that though). There are no issues for that so I don't think anyone actually uses the original gyp anymore?

joyeecheung:
cc @indutny @bnoordhuis @hhellyer  what do you think?

bnoordhuis:
I'm fine with that.

indutny:
:+1:

PR #182, opened 2018-03-30
build: merge gyp_llnode into scripts/configure.js

joyeecheung:
This patch drops the support for the original gyp. Building the
plugin now requires node-gyp which detects the necessary
configurations that we need from gyp_llnode. The original
gyp build flow has been broken without being noticed so it
should be safe to remove support for that.

- Renames the gyp and gypi files so they are easier to recognize
  - llnode.gyp -> binding.gyp
  - common.gypi -> llnode.gypi
  - options.gypi -> config.gypi
- Renames the gyp variables
  - lldb_dir to lldb_header_dir because that's the directory used
    to search for headers.
  - lldb_build_dir to lldb_lib_dir because that's the directory
    used to search for library files.
  - Added documentation on how to configure these variables from
    the command line.
- Clarify that the lldb_exe npm config is only used to retrieve
  lldb versions.

Refs: https://github.com/nodejs/llnode/issues/181


package.json
```diff
@@ -9,10 +9,9 @@
   "bin": {
     "llnode": "llnode.sh"
   },
-  "//": "(Blame C++)",
   "scripts": {
     "preinstall": "node scripts/configure.js",
-    "install": "./gyp_llnode && ( gmake -C out/ || make -C out/ )",
+    "install": "node-gyp rebuild",
```
richardlau: Is this required? npm will default to running `node-gyp` if it finds a `binding.gyp`.

joyeecheung: @richardlau It does not run that if there is a `preinstall` hook

Issue #187, opened 2018-04-24
Periodic test against master and canary


mmarchini:
I think we should have nightly/weekly test runs against master and canary. This would help us know earlier when something breaks. https://github.com/nodejs/node/pull/17685 was an awesome start, but as @bnoordhuis said in the PR it won't catch semantic changes.

I'm willing to setup the infrastructure for it. I just need to know if we can have this on Travis (by using https://docs.travis-ci.com/user/cron-jobs/) or if we should introduce this on Node.js CI.

/cc @joyeecheung

mmarchini:
We could even have a status table in the readme 😄 

| Version | Status  |
|---------|---|
| 6.x     | ![build passing](https://img.shields.io/badge/build-passing-brightgreen.svg)  |
| 8.x     | ![build passing](https://img.shields.io/badge/build-passing-brightgreen.svg)  |
| 9.x     | ![build passing](https://img.shields.io/badge/build-passing-brightgreen.svg)  |
| 10.x    | ![build passing](https://img.shields.io/badge/build-failing-red.svg)  |
| master  | ![build passing](https://img.shields.io/badge/build-failing-red.svg)  |
| canary  | ![build passing](https://img.shields.io/badge/build-failing-red.svg)  |

joyeecheung:
There is https://github.com/nodejs/build/pull/1025 by the way.

I have seen people wgetting nightlies in the scripts to run tests on nightlies with Travis, but that looks rather hacky...

mmarchini:
I saw that PR a while ago, but it seems stalled for some time. I'll take a look and see if we can get some traction on it.

BTW, should we move our current CI infrastructure from Travis to our Jenkins?

joyeecheung:
> BTW, should we move our current CI infrastructure from Travis to our Jenkins?

@mmarchini That's the plan, but until it happens Travis is better than nothing.

mmarchini:
> I have seen people wgetting nightlies in the scripts to run tests on nightlies with Travis, but that looks rather hacky...

I looked into Travis docs and it seems like they use nvm to manage Node.js versions. *Maybe* we can use it to get nightlies in a not-so-hacky way. I think I'll try to add something to Travis while we don't have support for llnode on Jenkins.

joyeecheung:
> Setting up Cron jobs is straightforward, but need admin access to Travis I think. It can be done in the menu Settings at the end of the page.

[...]

PR #189, opened 2018-05-02
travis: increase versions and OS coverage

mmarchini:
* Reworked travis.yaml to use matrix settings, giving us more control
  over each job settings.
* Remove v4.x, since it's not a supported release line anymore
* Add jobs v10.x
* Add jobs for Mac OS X
* Add jobs for Node.js master and v8-canary builds. Those jobs are
  ignored in the build final result.
* Add a build status table to the README

Ref: https://github.com/nodejs/llnode/issues/187

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,8 +40,6 @@ plugin: configure
 .PHONY: _travis
 _travis:
 	TEST_LLDB_BINARY="$(TEST_LLDB_BINARY)" \
-	TEST_LLNODE_DEBUG=true \
-	LLNODE_DEBUG=true \
 	npm test
 
 .PHONY: clean
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,82 @@
 language: node_js
-sudo: required
-dist: trusty
-before_install:
-  - sudo apt-get -qq update
-  - sudo apt-get install lldb-3.9 liblldb-3.9-dev -y
-node_js:
-  - "4"
-  - "6"
-  - "8"
-  - "9"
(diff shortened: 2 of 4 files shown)


mmarchini:
BTW this is how the README would look like with these changes:

image

mmarchini:
We might also need to increase timeouts for Mac OS X jobs (Travis Mac OS X machines are rather slow and I've got a lot of timeouts already), but this could be done in a follow-up PR.

mmarchini:
Apparently the `allow_failures` parameter is not working... Let me see if I can fix it before we land this

mmarchini:
> Apparently the allow_failures parameter is not working...

Fixed.

I've also increased the timeout to save and load the core file to avoid failures on OS X machines.

Makefile
```diff
@@ -40,8 +40,6 @@ plugin: configure
 .PHONY: _travis
 _travis:
 	TEST_LLDB_BINARY="$(TEST_LLDB_BINARY)" \
-	TEST_LLNODE_DEBUG=true \
-	LLNODE_DEBUG=true \
```
mmarchini: @joyeecheung I removed those lines because I think this might be affecting performance on OS X machines when generating a core file:

image

Besides, we have too much noise today with those flags on, especially with fallback symbols and "Not object and not smi", making it harder to read the log. If you want I can revert these changes though.

joyeecheung: @mmarchini No problems, I think we can enable them back in PRs that need them for debugging purposes.

Issue #158, opened 2017-12-19
Upcoming metadata changes


cjihrig:
The following changes are coming:

- `v8dbg_class_Map__inobject_properties_or_constructor_function_index__int`
  - This is now `v8dbg_class_Map__inobject_properties_start_or_constructor_function_index__char` as of https://github.com/v8/v8/commit/61bf2cc69217a4e9c7a40bb74269508fa26c2062
- `v8dbg_class_Map__instance_attributes__int`
  - This is now `v8dbg_class_Map__instance_type__uint16_t` as of https://github.com/v8/v8/commit/c00bb6da1cef9f7e9dab842904232255717f8291 and https://github.com/v8/v8/commit/cb46310a799eb0d60a613ead0af8c8200c5338b1
- `v8dbg_class_Map__instance_size__int`
  - This is now `v8dbg_class_Map__instance_size_in_words__char` as of https://github.com/v8/v8/commit/61bf2cc69217a4e9c7a40bb74269508fa26c2062
- `v8dbg_bit_field3_dictionary_map_shift`
  - This is now `v8dbg_bit_field3_is_dictionary_map_shift` as of https://github.com/v8/v8/commit/7a159da151e5f5437d6924de4fd142f854d3df4e

See https://github.com/nodejs/node-v8/pull/34

mmarchini:
These metadata changes landed with https://github.com/nodejs/node/pull/17489, thus llnode is not working with `node` compiled from source. There's also one more missing: `v8dbg_prop_type_mask`.

cjihrig:
> There's also one more missing: v8dbg_prop_type_mask.

That one is expected. It's used [here](https://github.com/nodejs/llnode/blob/20b14d967ecfe60a3b778611de04bad4f59b14b9/src/llv8-constants.cc#L461-L463) for Node.js version detection.

mmarchini:
I think we should reopen this Issue, we updated the metadata names, but we didn't fix semantic changes. `npm test` have six failing tests on Node.js v10.0.0.

I started to investigate them, and I think at least three failures are happening because of https://github.com/v8/v8/commit/61bf2cc69217a4e9c7a40bb74269508fa26c2062. If I understood the patch correctly, it moved the internal properties of `Map`, confusing llnode once it tries to inspect objects. For example, in our inspect-test, Class.x should be 1 and Class.hashmap should be an object, but instead, they both are pointing to the internal Map: 

```
(lldb) v8 findjsinstances -d Class
[...]

PR #192, opened 2018-05-03
src: fix object inspection for V8 6.4

mmarchini:
V8 6.4 "replaces the in-object properties count byte in the map
with the byte that stores the start offset of in-object properties".
Object inspection on llnode relies on in-object properties being count
bytes, so these are minimal changes to make object inspection work again
with V8 6.4 while keeping compatibility with previous versions.

Ref: https://chromium-review.googlesource.com/c/v8/v8/+/776720
Fixes: https://github.com/nodejs/llnode/issues/158

### DIFF
--- a/src/llv8-constants.cc
+++ b/src/llv8-constants.cc
@@ -209,6 +209,7 @@ void Map::Load() {
         "char");
   }
 
+  kInstanceTypeOffset = LoadConstant("class_Map__instance_type__uint16_t");
   kInstanceSizeOffset = LoadConstant("class_Map__instance_size__int",
                                      "class_Map__instance_size_in_words__char");
   kDictionaryMapShift = LoadConstant("bit_field3_dictionary_map_shift",
@@ -574,6 +575,8 @@ void Frame::Load() {
 
 void Types::Load() {
   kFirstNonstringType = LoadConstant("FirstNonstringType");
+  kFirstJSObjectType =
+      LoadConstant("type_JSGlobalObject__JS_GLOBAL_OBJECT_TYPE");
 
   kHeapNumberType = LoadConstant("type_HeapNumber__HEAP_NUMBER_TYPE");
   kMapType = LoadConstant("type_Map__MAP_TYPE");
--- a/src/llv8-constants.cc
+++ b/src/llv8-constants.cc
@@ -204,7 +204,7 @@ void Map::Load() {
       "class_Map__inobject_properties_or_constructor_function_index__int",
       "class_Map__inobject_properties__int");
   if (kInObjectPropertiesOffset == -1) {
-    kInObjectPropertiesOffset = LoadConstant(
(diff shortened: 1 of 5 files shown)


src/llv8-inl.h
```diff
@@ -160,7 +160,21 @@ inline int64_t Map::BitField3(Error& err) {
...
+  if (v8()->map()->kInObjectPropertiesOffset != -1) {
+    return LoadField(v8()->map()->kInObjectPropertiesOffset, err) & 0xff;
+  } else {
+    // NOTE(mmarchini): V8 6.4 changed semantics for
+    // in_objects_properties_offset (see
+    // https://chromium-review.googlesource.com/c/v8/v8/+/776720). To keep
+    // changes to a minimum on llnode and to comply with V8, we're using the
+    // same implementation from
+    // https://chromium-review.googlesource.com/c/v8/v8/+/776720/9/src/objects-inl.h#3027.
+    int64_t in_objects_start =
```
bnoordhuis: `in_object_properties_start_offset`?

src/llv8-inl.h
```diff
@@ -160,7 +160,21 @@ inline int64_t Map::BitField3(Error& err) {
...
+        LoadField(v8()->map()->kInObjectPropertiesStartOffset, err) & 0xff;
+    return v8()->LoadUnsigned(LeaField(v8()->map()->kInstanceSizeOffset), 1,
```
bnoordhuis: For legibility, maybe assign the result of LeaField() to a temporary first.

mmarchini:
> I thought it only applies when the map's instance type >= FIRST_JS_OBJECT_TYPE, i.e., when it's a map describing a JSObject.

There are 5 calls to `InObjectProperties` in llnode, 4 of them are in JSObject methods, so those calls should be safe. The only one I'm not 100% sure is correct is the call in `Map::Inspect`. We might need to do some special handling there, but I'm not sure. I'll leave a FIXME comment to investigate this in a follow-up PR, WDYT?

mmarchini:
Improved legibility on InObjectProperty and handled the case where Map don't describe a JSObject. @bnoordhuis PTAL.

Would be nice to have a test for this case. I'll see if I can come up with one later, but suggestions are appreciated 😄

src/llv8-constants.cc
```diff
@@ -574,6 +575,8 @@ void Frame::Load() {
 
 void Types::Load() {
   kFirstNonstringType = LoadConstant("FirstNonstringType");
+  kFirstJSObjectType =
+      LoadConstant("type_JSGlobalObject__JS_GLOBAL_OBJECT_TYPE");
```
cjihrig: Is this a new constant? Just asking, because if it is, I'd like to add it to Node's postmortem metadata test.

mmarchini: Yep

cjihrig: Looks like it's already part of the test.

Issue #197, opened 2018-05-08
Should we show "Class objects" as objects?


mmarchini:
* **Version**: llnode 1.7.0, node v8.11.1
* **Platform**: OS X (10.13.4)
* **Subsystem**: findjsobjects, findjsinstances, inspect

Right now if we declare a class (let's say, `Lira`), and we don't create an instance of that class, `v8 findjsobjects` will still show us that we have 1 instance of `Lira`. If we have a class with instances (let's say, `Lero`), `v8 findjsobjects` will show us one more instance than we actually have. Example of this behavior below.

I'm not sure how mdb_v8 handles this, but Chrome DevTools snapshots won't count classes as objects. Should we change the current behavior? Or maybe make it clearer when some object is actually a class instead of an instance?

Example:

**index.js**
```js
class Lero {
  constructor() {
    this.lala = {};
  }
}

class Lira {
  constructor() {
    this.lele = {};
  }
}


const lero = new Lero();

function foo() {
  const lero2 = new Lero();
  boom();
}

foo();
```

**llnode results**
```
 Instances  Total Size Name
 ---------- ---------- ----
          1         24 AssertionError
          1         24 AsyncResource
          1         24 FastBuffer
          1         24 Lira
          1         24 Loader
          1         24 ModuleJob
          1         24 ModuleMap
          1         24 Performance
          1         24 PerformanceObserver
          1         24 SafeMap
          1         24 SafePromise
          1         24 SafeSet
          1         24 TextDecoder
          1         24 TextEncoder
          1         24 TickObject
          1         24 URL
          1         24 URLContext
          1         24 URLSearchParams
          1         24 WebAssembly
          1         24 console
          1         32 (Object)
          1        104 ImmediateList
          1        104 Stack
          1        112 Console
          2         48 (anonymous)
          2         48 process
          2         64 ChannelWrap
          2         64 Signal
          2        120 Resolver
          2        128 PerformanceNodeTiming
          2        136 NextTickQueue
          2        200 PerformanceObserverEntryList
          2        208 WriteStream
          2        272 Module
          2        608 ReadableState
          2        640 WritableState
          3         72 NodeError
          3         96 TTY
          3        216 Lero
          3        248 BufferList
          3        280 AsyncHook
         12        384 ContextifyScript
         14       1120 (ArrayBufferView)
         55       3520 NativeModule
        327      10464 (Array)
        642      36248 Object
       6896      53488 (String)
 ---------- ----------
       8009     109504
```

[...]

Issue #195, opened 2018-05-08
Should findrefs display context/scope objects?


mmarchini:
* **Version**: llnode 1.7.0, node v8.11.1
* **Platform**: OS X (10.13.4)
* **Subsystem**: findrefs

Let's say we have an object which is not referenced by any other objects, but is scoped by a function or by a script. Should `v8 findrefs [obj addr]` return nothing (current behavior), or should it return the scope this object belongs to? IMO we should change the behavior to return the scope as well, because looking at the output from the example below it seems like there are `Lero` instances leaking, even though it's not true.

Example:

**index.js**
```js
class Lero {
  constructor() {
    this.lala = {};
  }
}

class Lira {
  constructor() {
    this.lele = {};
  }
}


const lero = new Lero();

function foo() {
  const lero2 = new Lero();
  boom();
}

foo();
```

**llnode results**
```
(llnode) v8 findjsinstances -d Lero
0x00000548a34200e1:<Object: Lero properties {
    .constructor=<unknown field type>}>
0x00000548a3420589:<Object: Lero properties {
    .lala=0x00000548a34205e9:<Object: Object>}>
0x00000548a3420659:<Object: Lero properties {
    .lala=0x00000548a34206b9:<Object: Object>}>
(llnode) v8 findrefs 0x00000548a3420589
(llnode) v8 findrefs 0x00000548a3420659
```

bnoordhuis:
Including the owning context(s) in the output sounds like a good idea to me.

joyeecheung:
SGTM, otherwise one will hit a dead end with objects like that.

Drieger:
A just created a PR [llnode/pull/227](https://github.com/nodejs/llnode/pull/227) to approach this issue. It seems that some Function-Variables are not referenced by any `Context` unless a nested closure references it.

The idea is to store all contexts and look in each one when we try to find references for an object.

watson:
Wasn't this addressed in #227?

mmarchini:
Yes. I was keeping this open as a reminder to show stack locals in findrefs, but I'm not even sure it's possible. Closing.

PR #199, opened 2018-05-16
[WIP] src: `v8 snapshot` command

mmarchini:
**DO NOT LAND YET**

[P.S.: The code is a mess right now, I'm opening a PR to get feedback on how the feature should look like]

This PR introduces a new command to create a V8 snapshot from a core dump. Right now the snapshot is not fully functional, but it works on Chrome DevTools. The goal is to enable the use of Chrome DevTools alongside llnode.

We still need to address https://github.com/nodejs/llnode/issues/195 and https://github.com/nodejs/llnode/issues/197 before making it work properly. Right now it can be used with the command `v8 snapshot`, which will create a `core.heapsnapshot` in the current working directory. The generated file can be open by Chrome DevTools, but it doesn't show a list of retainers yet.

### DIFF
--- a/src/llnode.cc
+++ b/src/llnode.cc
@@ -2,6 +2,8 @@
 #include <getopt.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sstream>
+#include <fstream>
 
 #include <cinttypes>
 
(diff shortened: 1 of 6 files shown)


indutny:
Neat! Just in case, you may find this helpful: https://github.com/indutny/core2dump (if it is still relevant)

mmarchini:
I found https://github.com/indutny/core2dump when working on this, but couldn't make it work :( (probably because so many things have changed in V8 in the last two years)

joyeecheung:
It would be cool to be able to expose a `toHeapSnapshot()` method in https://github.com/nodejs/llnode/pull/147 I plan to revamp that PR a bit soon to expose structured objects instead of strings.

mmarchini:
Closing since I haven't had the time to move this forward. If anyone wants to work on it, feel free to pick it up! (Although I think this will be much easier to accomplish once our JS API is out of experimental).

Issue #17, opened 2016-07-12
LLNode Windows support


hhellyer:
I thought I’d open this issue to discuss whether Windows support for LLNode is something anyone wants. 

I’ve got and LLDB build up and running on Windows so I had a chance to try building LLNode on Windows. Running gyp_llnode worked fairly well.

The resulting Visual Studio project created by gyp required a little tweaking (mostly things I could google for). The main issue was that LLNode uses the getopt library which isn’t on Windows by default. Given that the getopt problem could probably be fixed running LLNode on Windows looks do-able but would need a bit of work and would probably break fairly quickly unless someone was using it regularly.

There’s two reasons why someone might want to do this:

- To debug issues with versions of Node running on Windows.
- To debug issues with core dumps generated on Linux. (LLDB is able to open Linux ELF cores even when it’s running on Mac or Windows.)

I’m not sure how important the first scenario is but the second one might be useful to developers with Windows machines who deploy to Linux. Of course you do have to get LLDB for Windows, probably by building it yourself, which isn’t hard but it is quite time consuming. (Developers with Mac’s who deploy to Linux are effectively already supported.)

Any thoughts? Does anyone else have a strong need for this? (I’m using a Mac so I’m neutral.)

indutny:
I think this is good addition, but we will probably need to hook it up to node.js CI, so that it will always be tested too. cc @rvagg @Fishrock123

rvagg:
Opened an issue here to discuss the question of how a project might qualify to use our infra: https://github.com/nodejs/build/issues/448, I think it might be a tough call because it'll open floodgates for projects not clearly within the scope of the foundation.

indutny:
I don't think there is any reason for llnode to not move under Node.js Foundation. If this is a show stopper - what could be done to make this project qualify for this?

rvagg:
I have no problem with llnode to move into nodejs/, perhaps raise an issue in the diagnostics WG and ask if they'd like to take some broad responsibility for it?

indutny:
@rvagg absolutely?

@hhellyer will this scenario work for you, or would you prefer the project to remain independent?

indutny:
@rvagg that was "!" not "?", sorry for confusion.

hhellyer:
I think the moving the project under Node.js would be great, it would probably improve the visibility and hopefully adoption and participation as well.
I'm not sure I know enough about the Node.js build infrastructure for that part of this discussion but I can see why they'd be worried about the load from adding various other projects.

[...]

PR #203, opened 2018-06-07
Add Windows Support

joaocgreis:
This PR adds support to build llnode on Windows and load core files generated on other platforms. Debugging the Windows Node executable is still not working. This would be nice to have as a base to keep developing Windows support. However, if this is considered too green to land, it can serve as a straw-man for discussion and feedback is welcome.

This makes two relevant changes that are not restricted to Windows:
- Convert the llnode shortcut to javascript, to make it cross-platform. It is not possible to add a `llnode.cmd` script because npm already creates that file pointing to `llnode.sh`.
- Add an install script, instead of calling `node-gyp` directly. This is necessary to compile the module using Clang instead of CL.

Because this is not working to debug a live Windows executable, all tests fail. Still, it is possible to run `scan-test.js` with a core file generated on another platform as [described in the README](https://github.com/nodejs/llnode#useful-environment-variables). The core file must be generated using a 32bit node executable as lldb is [not ready for 64bit usage on Windows](https://lldb.llvm.org/status.html).

Ref: https://github.com/nodejs/llnode/issues/17


scripts/cleanup.js
```diff
@@ -2,8 +2,14 @@
 
 const fs = require('fs');
 const cwd = process.cwd();
+const osName = require('os').type();
 
-const libExt = require('os').type() === 'Darwin' ? 'dylib' : 'so';
+let libExt = 'so';
```
cjihrig: Can you just make `'so'` an `else` case.

joaocgreis:
@cjihrig thanks, updated! Let me know if this is not what you had in mind.

joaocgreis:
Rebased, applied the fixup and added a `PR-URL` (previous iteration in https://github.com/joaocgreis/llnode/compare/master...joaocgreis-I5O-windows-support-0).

I understand that more reviews might be necessary before this can land, but when the time comes I won't be able to do it myself as I don't have write access in this repo.

joyeecheung:
It would nice if we can have some appveyor coverage for this, but I think it's ready to land. Thanks for the work on this!

PR #204, opened 2018-06-13
src: add node's internals constants

mmarchini:
This commit refactors llv8-constants to make it easier to introduce
Node's internals constants. Common code for llv8 and node constants is
now on src/constants.h. Also moved the Error class to its own file,
removing the dependency on src/llv8.h to use Errors.

### DIFF
--- a/src/node-constants.cc
+++ b/src/node-constants.cc
@@ -1,0 +1,137 @@
+#include <lldb/API/LLDB.h>
+#include <set>
+
+#include "src/llv8-inl.h"
+#include "src/node-constants.h"
+
+using lldb::SBFrame;
+using lldb::SBProcess;
+using lldb::SBStream;
+using lldb::SBThread;
+
+namespace llnode {
+namespace node {
+namespace constants {
+
+void Environment::Load() {
+  kReqWrapQueueOffset = LoadConstant(
+      "offset_Environment__req_wrap_queue___Environment_ReqWrapQueue");
+  kHandleWrapQueueOffset = LoadConstant(
+      "offset_Environment__handle_wrap_queue___Environment_HandleWrapQueue");
+  kEnvContextEmbedderDataIndex =
+      LoadConstant("const_Environment__kContextEmbedderDataIndex__int");
+
+  Error err;
+  kCurrentEnvironment = LoadCurrentEnvironment(err);
+}
+
+addr_t Environment::LoadCurrentEnvironment(Error& err) {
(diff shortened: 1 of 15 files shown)


mmarchini:
Split from #122 as I think it will be easier to review these changes in separate PRs.

src/node-constants.cc
```diff
@@ -0,0 +1,152 @@
...
+  SBProcess process = target_.GetProcess();
+  SBThread thread = process.GetSelectedThread();
+  if (!thread.IsValid()) {
+    err = Error::Failure("Invalid thread");
+    return 0;
+  }
+
+  llv8()->Load(target_);
+
+  SBStream desc;
```
joyeecheung: Are `desc` and `selected_frame` used anywhere?

src/node-constants.cc
```diff
@@ -0,0 +1,152 @@
...
+      bool found = false;
+      std::set<uint64_t> visited_contexts;
+      while (!found) {
+        // Necessary to avoid an infinite loop.
+        if (visited_contexts.count(val.raw())) {
+          break;
+        }
+        visited_contexts.insert(val.raw());
+        v8::Context context(val);
+        v8::Value native = context.Native(v8_err);
```
joyeecheung: It would be cleaner if we put this into a `Context::IsNativeConext()`

src/node-constants.cc
```diff
@@ -0,0 +1,152 @@
...
+        if (v8_err.Success()) {
+          if (native.raw() == context.raw()) {
```
joyeecheung: This works, but V8 seems to prefer comparing the map with `kNativeContextMapRootIndex` (not yet available in the post-mortem constants, it seems)

src/node-constants.cc
```diff
@@ -0,0 +1,152 @@
...
+
+  return currentEnvironment;
+}
+
+addr_t Environment::CurrentEnvironmentFromContext(v8::Value context,
+                                                  Error& err) {
+  llv8()->Load(target_);
+
+  v8::FixedArray contextArray = v8::FixedArray(context);
+  v8::FixedArray embed = contextArray.Get<v8::FixedArray>(
```
joyeecheung: It may be cleaner if we wrap this into a `Context::GetEmbedderData(index)`

mmarchini:
If there are no objections I'd like to land this by the end of the week.

mmarchini:
Landed in 3c3fb94

PR #205, opened 2018-06-21
src: inspect unicode string correctly

hyj1991:
image

### DIFF
--- a/src/llv8.cc
+++ b/src/llv8.cc
@@ -200,6 +200,57 @@ std::string LLV8::LoadString(int64_t addr, int64_t length, Error& err) {
   return res;
 }
 
+std::string LLV8::Utf16ToUtf8(const std::u16string& u16_str) {
+  if (u16_str.empty()) {
+    return std::string();
+  }
+  const char16_t* p = u16_str.data();
+  std::u16string::size_type len = u16_str.length();
+  if (p[0] == 0xFEFF) {
+    // bom
+    p += 1;
+    len -= 1;
+  }
+
+  std::string u8_str;
+  u8_str.reserve(len * 3);
+
+  char16_t u16_char;
+  for (std::u16string::size_type i = 0; i < len; ++i) {
+    u16_char = p[i];
+    if (u16_char < 0x0080) {
+      u8_str.push_back((char)(u16_char & 0x00FF));
+      continue;
+    }
+    if (u16_char >= 0x0080 && u16_char <= 0x07FF) {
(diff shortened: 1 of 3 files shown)


src/llv8.cc
```diff
@@ -200,14 +200,65 @@ std::string LLV8::LoadString(int64_t addr, int64_t length, Error& err) {
...
+    {
+      u8_str.push_back((char)(((u16_char >> 12) & 0x0F) | 0xE0));
+      u8_str.push_back((char)(((u16_char >> 6) & 0x3F) | 0x80));
+      u8_str.push_back((char)((u16_char & 0x3F) | 0x80));
+      continue;
+    }
+  }
+
+  return u8_str;
+}
```
bnoordhuis: Can this be replaced by `ConvertUTF16toUTF8()` from llvm's `Support/ConvertUTF.h`?

src/llv8.cc
```diff
@@ -220,10 +271,9 @@ std::string LLV8::LoadTwoByteString(int64_t addr, int64_t length, Error& err) {
     return std::string();
   }
 
-  for (int64_t i = 0; i < length; i++) buf[i] = buf[i * 2];
-  buf[length] = '\0';
+  char16_t* u16str = (char16_t*) buf;
```
bnoordhuis: This isn't safe, unfortunately, because it aliases and `buf` need not be properly aligned (although it probably is since it was new'd.)

Wouldn't calling `new char16_t[length]` work?  Since you're here, maybe switch to `std::vector<char16_t>`.

Aside: `ConvertUTF16toUTF8()` uses `llvm::UTF16` (typedef for `unsigned short`), not `char16_t`.

src/llv8.cc
```diff
@@ -1047,10 +1097,10 @@ std::string String::Inspect(InspectOptions* options, Error& err) {
   if (err.Fail()) return std::string();
 
   unsigned int len = options->length;
-
+  int total_length = val.length();
   if (len != 0 && val.length() > len) val = val.substr(0, len) + "...";
 
-  return "<String: \"" + val + "\">";
+  return "<String: \"" + val + "\", length=" + std::to_string(total_length) + ">";
```
joyeecheung: Can you make this a command line option? Or display length before the content like what `JSArray` does.

hyj1991: ok

src/llv8.cc
hyj1991: thanks for reviewing, I'll fix this

hyj1991:
new pr: https://github.com/nodejs/llnode/pull/207

PR #207, opened 2018-06-25
src: inspect unicode string correctly

hyj1991:


src/llv8.cc
```diff
@@ -200,14 +200,158 @@ std::string LLV8::LoadString(int64_t addr, int64_t length, Error& err) {
   return res;
 }
 
+void LLV8::ConvertUTF16toUTF8 (
+  const UTF16** sourceStart, const UTF16* sourceEnd, UTF8** targetStart,
```
joyeecheung: If you are copying code from llvm, can you split these functions out to a separate file and add the attribution required in the header of https://github.com/llvm-mirror/llvm/blob/master/lib/Support/ConvertUTF.cpp ?

Alternatively you can include headers from llvm and link to their shared library but I think that would be a much bigger hassle in terms of changes to our build files.

hyj1991: oh...didn't notice attention to the issue of open source agreements...

PR #210, opened 2018-07-01
 src: add commands to inspect the workqueue

mmarchini:
Add two new commands: `v8 getactivehandles` and `v8 getactiverequests`.
These commands will print all pending handles and requests. The result
should be similar to running process._getActiveHandles() and
process._getActiveRequests() on the living process.

Fixes: #100


mmarchini:
Landed in baf4cae

Issue #156, opened 2017-12-08
inspect objects with "Symbol" attributes


mmarchini:
Inspect commands can't handle Symbol attributes (e.g. @@toStringTag and @@unscopable), resulting in `<non-string>` properties. 

For example, the following code uses the `@@toStringTag` symbol:

```javascript
class Bar {
  constructor() {
    this[Symbol.toStringTag] = "Bar";
  }
}

let bar = new Bar();
```

If we use llnode to find out the properties for `bar`, it will give the following output:

```
(lldb) v8 findjsinstances -v Bar
0x0000015078184729:<Object: Bar properties {
    .constructor=0x00000150781846e1:<function: Bar at index.js:17:14>}>
0x0000015078184949:<Object: Bar properties {
    .<non-string>=0x00002d2064acb1e1:<String: "Bar">}>
```

As you can see, there's a `<non-string>` property instead of our symbol.

I was looking at llnode and V8 code to try to find a solution, but I'm not sure how to implement it. Did anyone try this before?

bnoordhuis:
Can you try this patch with node.js master?  That should add the metadata for symbols although llnode still needs to be taught to use it.
```diff
diff --git a/deps/v8/src/v8.gyp b/deps/v8/src/v8.gyp
index 89eb271f61..38670c193d 100644
--- a/deps/v8/src/v8.gyp
+++ b/deps/v8/src/v8.gyp
@@ -2452,6 +2452,8 @@
             'objects-inl.h',
             'objects/map.h',
             'objects/map-inl.h',
+            'objects/name.h',
+            'objects/name-inl.h',
             'objects/script.h',
             'objects/script-inl.h',
             'objects/shared-function-info.h',
diff --git a/deps/v8/tools/gen-postmortem-metadata.py b/deps/v8/tools/gen-postmortem-metadata.py
index 22f0afbef3..5f33fdebe9 100644
--- a/deps/v8/tools/gen-postmortem-metadata.py
+++ b/deps/v8/tools/gen-postmortem-metadata.py
@@ -417,15 +417,10 @@ def load_objects_from_file(objfilename, checktypes):
         # way around.
         #
         for type in types:
-                #
-                # Symbols and Strings are implemented using the same classes.
-                #
-                usetype = re.sub('SYMBOL_', 'STRING_', type);
-
                 #
                 # REGEXP behaves like REG_EXP, as in JS_REGEXP_TYPE => JSRegExp.
                 #
[...]

PR #214, opened 2018-07-12
src: teach llnode how to print Symbol properties names

mmarchini:
Still need to add the metadata to V8 before landing this.

Closes: https://github.com/nodejs/llnode/issues/156

### DIFF
--- a/src/llv8.cc
+++ b/src/llv8.cc
@@ -60,6 +60,7 @@ void LLV8::Load(SBTarget target) {
   descriptor_array.Assign(target, &common);
   name_dictionary.Assign(target, &common);
   frame.Assign(target, &common);
+  symbol.Assign(target, &common);
   types.Assign(target, &common);
 }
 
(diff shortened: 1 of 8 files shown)


src/llv8.cc
```diff
@@ -955,6 +961,15 @@ std::string HeapNumber::Inspect(Error& err) {
 }
 
 
+std::string Symbol::ToString(Error& err) {
+  if (!String::IsString(v8(), Name(err), err)) {
```
cjihrig: Any reason not to do `Symbol()`/`Symbol('foo')`, like would be shown in JS?

mmarchini: You mean for unnamed symbols? `Symbol()` sounds like a good idea in this case.

mmarchini:
Rebased, changed display format to `Symbol()/Symbol('foo')` and added code to fallback to `<non-string>` when Symbols metadata is not available (Node.js 6 and 8).

mmarchini:
Landed in e30beba

PR #219, opened 2018-08-08
build, doc: drop support for Node.js v9.x

mmarchini:
Since v9.x is EOL, we can remove it from our CI matrix. I also added names to our Travis jobs and reorganized `.travis.yml` (inspired on https://github.com/elastic/apm-agent-nodejs/blob/09316fa7c33330b7264d376ab219e85dd384a97a/.travis.yml#L184-L264).


mmarchini:
Landed in 5b6a6b5...65debbd

Issue #198, opened 2018-05-10
Methods appearing as "<unknown field type>" on Node.js v8.x+


mmarchini:
* **Version**: llnode master
* **Platform**: OS X (10.13.4)
* **Subsystem**: findjsinstances -d, inspect

llnode can't inspect methods on Node.js v8.x or later. v6.x works fine.

**node v6.14.2**
```
(lldb) v8 findjsinstances -d Lero
0x00001897d2db3639:<Object: Lero properties {
    .constructor=0x00001897d2db35d1:<function: Lero at /Users/mmarchini/tmp/llnode-example/index.js:4:14>,
    .foo=0x00001897d2db3841:<function: foo at /Users/mmarchini/tmp/llnode-example/index.js:8:6>}>
0x00001897d2db3fb9:<Object: Lero properties {
    .lala=0x00001897d2db4051:<Object: Object>}>
0x00001897d2db40d1:<Object: Lero properties {
    .lala=0x00001897d2db4131:<Object: Object>}>
```

**node v8.0.0** (or later)
```
(lldb) v8 findjsinstances -d Lero
0x000008c85fc57999:<Object: Lero properties {
    .constructor=<unknown field type>,
    .foo=<unknown field type>}>
0x000008c85fc58161:<Object: Lero properties {
    .lala=0x000008c85fc581c1:<Object: Object>}>
0x000008c85fc58241:<Object: Lero properties {
    .lala=0x000008c85fc582a1:<Object: Object>}>
```

**index.js**
```js
class Lero {
  constructor() {
    this.lala = {};
  }

  foo(a) {
    return a;
  }
}

class Lira {
  constructor() {
    this.lele = {};
  }
}

function Lura() {
  this.lele = 1;
  this.constructor = function() {};
}

const lura = new Lura();
lura.lorem = "ipsum";

const lero = new Lero();

function foo() {
  const lero2 = new Lero();
  boom();
}


foo();
```

Issue #209, opened 2018-06-28
process object with weird attributes in Node.js v10.5.0


mmarchini:
```
0x000002e56df83d09:<Object: process properties {
    .title=<unknown field type>,
    .version=<Smi: 1>,
    .versions=<Smi: 1>,
    .arch=<Smi: 1>,
    .platform=<Smi: 1>,
    .release=<Smi: 1>,
    .argv=0x000002e5ff5aa5e9:<Array: length=2>,
    .execArgv=0x000002e5ff5aa609:<Array: length=2>,
    .env=0x000002e5ff5aa629:<Object: (anonymous)>,
    .pid=<Smi: 1>,
    .features=<Smi: 1>,
    .ppid=<unknown field type>,
    .execPath=0x000002e5ff5aa679:<String: "/Users/mmarchini...">,
    .debugPort=<unknown field type>,
    ._startProfilerIdleNotifier=<unknown field type>,
    ._stopProfilerIdleNotifier=<unknown field type>,
    .abort=<unknown field type>,
    .chdir=0x000002e51d4dbae1:<function: chdir at (external).js:20:33>,
    .umask=0x000002e51d4dbb21:<function: umask at (external).js:24:33>,
    ._getActiveRequests=<unknown field type>,
    ._getActiveHandles=<unknown field type>,
    .reallyExit=<unknown field type>,
    .cwd=<unknown field type>,
    .getuid=<unknown field type>,
    .geteuid=<unknown field type>,
    .getgid=<unknown field type>,
    .getegid=<unknown field type>,
    .getgroups=<unknown field type>,
    ._kill=<unknown field type>,
    .dlopen=<unknown field type>,
    ._debugProcess=<unknown field type>,
    ._debugEnd=<unknown field type>,
    .hrtime=0x000002e51d4dd1a1:<function: hrtime at (external).js:96:35>,
    .cpuUsage=0x000002e51d4dd4f9:<function: cpuUsage at (external).js:35:39>,
    .uptime=<unknown field type>,
    .memoryUsage=0x000002e51d4dd7a9:<function: memoryUsage at (external).js:123:45>,
    ._rawDebug=0x000002e51d4e0fd1:<function: process._rawDebug at (external).js:250:31>,
    .moduleLoadList=<Smi: 1>,
    .binding=<unknown field type>,
    ._linkedBinding=<unknown field type>,
    ._events=0x000002e5ff5aa6c9:<Object: Object>,
    ._eventsCount=<Smi: 4>,
[...]

PR #221, opened 2018-08-11
src: inspect properties with descriptor details

mmarchini:
When the PropertyLocation of the details of a JSObject property is
kDescriptor, we can just get the descriptor from the descriptors array
and inspect it. This solves several fields being resolved as
`<unknown field type>`.

Fixes: https://github.com/nodejs/llnode/issues/198
Ref: https://github.com/nodejs/llnode/issues/209

### DIFF
--- a/src/llv8-inl.h
+++ b/src/llv8-inl.h
@@ -510,10 +510,7 @@ inline bool DescriptorArray::IsConstFieldDetails(Smi details) {
   }
 
   // node.js >= 8
-  return (details.GetValue() &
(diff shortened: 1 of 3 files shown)


mmarchini:
Added another commit to address the `<Smi: 1>` properties issue. Our CI should be green now 😄

mmarchini:
I'll land this tomorrow if there are no objections.

mmarchini:
Landed in 65debbd...d348eff

PR #222, opened 2018-08-12
test: increase Linux test coverage

mmarchini:
Some tests (`test/addon/jsapi-test.js` and `test/plugin/scan-test.js`)
were not running on Linux because LLDB is not able to save core dumps on
that platform. These changes use a simpler approach to save core dumps
on Linux so these tests can also be run there.

Ref: https://github.com/nodejs/llnode/issues/209

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -205,10 +205,7 @@ Session.create = function create(scenario) {
   return new Session({ scenario: scenario });
 };
 
-exports.saveCore = function saveCore(options, cb) {
-  const scenario = options.scenario;
-  const core = options.core || exports.core;
-
+function saveCoreLLDB(scenario, core, cb) {
   // Create a core and test
   const sess = Session.create(scenario);
   sess.timeoutAfter(exports.saveCoreTimeout);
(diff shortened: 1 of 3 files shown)


mmarchini:
@joyeecheung thank you for your suggestions! I updated the PR to address them, and I also made a change to use `...; mv ./core ${core}` inside the command instead of `fs.rename` to avoid [a known issue with rename and docker containers](https://github.com/npm/npm/issues/9863#issuecomment-152342020).

mmarchini:
Ping @joyeecheung

test/common.js
```diff
@@ -234,6 +232,41 @@ exports.saveCore = function saveCore(options, cb) {
   });
 }
 
+function saveCoreLinux(executable, scenario, core, cb) {
+  const cmd = `ulimit -c unlimited && ${executable} ` +
+              `--abort_on_uncaught_exception --expose_externalize_string ` +
+              `${path.join(exports.fixturesDir, scenario)}; ` +
+              `mv ./core ${core}`;
```
joyeecheung: Can you split this into a separate command? It is possible that the core is not `./core` depending on the configuration of the system

mmarchini: Splitted into a separate command, but I think we should handle different configurations on a separate PR.

mmarchini:
Landed in 5cf24d0

PR #224, opened 2018-08-19
Load lldb symbols at runtime

springmeyer:
By using `-Wl,-undefined,dynamic_lookup` on OS X we can mirror the default behavior of `ld` on linux whereby undefined symbols in LLDB are found dynamically at runtime.

Using this flag allows the plugin to be created without linking error and without linking explicitly to the lldb shared library. The symbols are found when `plugin load <path to plugin>` is run within `lldb`.


joyeecheung:
@mmarchini Yeah moving the old config to the `build_addons` block SGTM

springmeyer:
What exactly needs to happen to get this merged?

> Yeah moving the old config to the build_addons block SGTM

Is this a change I need to make?

joyeecheung:
@springmeyer I think applying the diff in https://github.com/nodejs/llnode/pull/224#issuecomment-420292074 should make the test pass? The addon target needs the old config to statically linked to the shared library at build time.

springmeyer:
Thanks @joyeecheung - I'm understanding now. Just was not clicking before. I'll attempt this. In the meantime I noticed that it is only a single test failing. So, locally for me if I comment out just these lines then all the tests pass:

```diff
diff --git a/test/addon/jsapi-test.js b/test/addon/jsapi-test.js
index d067d37..95fe482 100644
--- a/test/addon/jsapi-test.js
+++ b/test/addon/jsapi-test.js
@@ -22,23 +22,23 @@ tape('llnode API', (t) => {
       t.error(err);
       t.ok(true, 'Saved core');
 
-      test(process.execPath, common.core, t);
+      //test(process.execPath, common.core, t);
       t.end();
     });
   }
 });
 
-function test(executable, core, t) {
-  debug('============= Loading ==============');
-  // Equivalent to lldb executable -c core
-  debug(`Loading core dump: ${core}, executable: ${executable}`);
-  const llnode = fromCoredump(core, executable);
-
-  verifySBProcess(llnode, t);
-  const typeMap = verifyBasicTypes(llnode, t);
-  const processType = verifyProcessType(typeMap, llnode, t);
-  verifyProcessInstances(processType, llnode, t);
-}
+// function test(executable, core, t) {
+//   debug('============= Loading ==============');
+//   // Equivalent to lldb executable -c core
+//   debug(`Loading core dump: ${core}, executable: ${executable}`);
+//   const llnode = fromCoredump(core, executable);
+//
+//   verifySBProcess(llnode, t);
+//   const typeMap = verifyBasicTypes(llnode, t);
+//   const processType = verifyProcessType(typeMap, llnode, t);
+//   verifyProcessInstances(processType, llnode, t);
+// }
 
 function verifySBProcess(llnode, t) {
   const processInfo = llnode.getProcessInfo();
```

springmeyer:
Okay, c08b3ac2404e implements the fix to:

 - continue using `-Wl,-undefined,dynamic_lookup`
 - but also keep the experiemental addon API working by direct linking, on OSX, to the LLDB frameowork, but just for the addon not the `shared_library`

PR #225, opened 2018-08-27
src: colorize output

Drieger:
Add colors to llnode output and introduce a settings class, which allow users to select if it should `always` use colorization, `never` use colorization or if it should `auto` select based on current output and color support. By default option `auto`, to change this option one should use the following command 
```
v8 settings set color <always|never|auto>
```
Almost all output of `findjsinstances` are already colorized, I'm still working on other commands. Image below shows an output example.

screenshot from 2018-08-27 10-52-29


src/llscan.cc
```diff
@@ -241,7 +233,19 @@ bool FindInstancesCmd::DoExecute(SBDebugger d, char** cmd,
     }
 
   } else {
-    result.Printf("No objects found with type name %s\n", type_name.c_str());
+    Settings* settings = Settings::GetSettings();
+    if (settings->ShouldUseColor()) {
+      rang::setControlMode(rang::control::Force);
+    } else {
+      rang::setControlMode(rang::control::Off);
+    }
```
mmarchini: This code is used frequently. It should probably be converted into a MACRO or inlined function. Or maybe we can call "rang::setControlMode" on SetPropertyColorCmd::DoExecute? I'm guessing this would be persistent across other commands, and also call it on `llnode::PluginInitialize`.

src/llscan.cc
```diff
@@ -241,7 +233,19 @@ bool FindInstancesCmd::DoExecute(SBDebugger d, char** cmd,
...
+
+    std::stringstream ss;
```
mmarchini: Suggestion: add a comment with the string format. For example:

```cpp
// "No objects found with type name %s", type_name
std::stringstream;
```

src/llscan.cc
```diff
@@ -241,7 +233,19 @@ bool FindInstancesCmd::DoExecute(SBDebugger d, char** cmd,
...
+    ss << rang::style::bold << rang::fg::red
+       << "No objects found with type name " << type_name.c_str()
```
mmarchini: I think `<< type_name` (without the `.c_str()`) will work, here and in other stringstreams.

src/llv8.cc
```diff
@@ -385,8 +392,19 @@ std::string JSFunction::Inspect(InspectOptions* options, Error& err) {
...
+    ss.clear();
+    ss << rang::style::bold << rang::fg::yellow << "\n  context"
+       << rang::fg::reset << rang::style::reset << "=" << rang::fg::cyan
+       << "0x%016" PRIx64 << rang::fg::reset;
+
+
     char tmp[128];
-    snprintf(tmp, sizeof(tmp), "\n  context=0x%016" PRIx64, context.raw());
+    snprintf(tmp, sizeof(tmp), ss.str().c_str(), context.raw());
+    // snprintf(tmp, sizeof(tmp), "\n  context=0x%016" PRIx64, context.raw());
```
mmarchini: Please remove commented code.

src/llv8.cc
```diff
@@ -385,8 +392,19 @@ std::string JSFunction::Inspect(InspectOptions* options, Error& err) {
 
     Context context(context_obj);
 
+    ss << rang::fg::magenta << res << rang::fg::reset;
+    res = ss.str();
```
mmarchini: I would rather have scoped stringstreams instead of clearing it every time (there are other places this pattern is used if you decide to change it):

```
  {
    std::stringstream ss;
    ss << rang::fg::magenta << res << rang::fg::reset;
    res = ss.str();
  }
```

src/llv8.cc
```diff
@@ -415,9 +433,13 @@ std::string JSFunction::Inspect(InspectOptions* options, Error& err) {
         res += source + "\n";
       }
     }
+    res = res + ">";
+  } else {
+    std::stringstream ss;
+    ss << rang::fg::yellow << res << ">" << rang::fg::reset;
+    res = ss.str().c_str();
```
mmarchini: At line 396 you called `res = ss.str();` (without `.c_str()`). I guess this should work here as well.

src/llv8.cc
```diff
@@ -854,7 +906,10 @@ std::string HeapObject::Inspect(InspectOptions* options, Error& err) {
 
   Error::PrintInDebugMode(
       "Unknown HeapObject Type %" PRId64 " at 0x%016" PRIx64 "", type, raw());
-  return pre + "<unknown>";
+
+  std::stringstream ss;
+  ss << rang::fg::yellow << "<unknown>" << rang::fg::reset;
+  return pre + ss.str().c_str();
```
mmarchini: `pre + ss.str()` should be enough.

src/llv8.cc
```diff
@@ -949,7 +1004,11 @@ std::string HeapObject::GetTypeName(Error& err) {
 }
 
 
-std::string Smi::Inspect(Error& err) { return "<Smi: " + ToString(err) + ">"; }
+std::string Smi::Inspect(Error& err) {
+  std::stringstream ss;
+  ss << rang::fg::yellow << "<Smi: " + ToString(err) + ">" << rang::fg::reset;
+  return ss.str().c_str();
```
mmarchini: ditto

src/llv8.cc
```diff
@@ -962,7 +1021,10 @@ std::string HeapNumber::ToString(bool whole, Error& err) {
 
 
 std::string HeapNumber::Inspect(Error& err) {
-  return "<Number: " + ToString(true, err) + ">";
+  std::stringstream ss;
+  ss << rang::fg::yellow << "<Number: " + ToString(true, err) + ">"
+     << rang::fg::reset;
+  return ss.str().c_str();
```
mmarchini: ditto

src/llv8.cc
```diff
@@ -1019,7 +1081,9 @@ std::string String::Inspect(InspectOptions* options, Error& err) {
 
   if (len != 0 && val.length() > len) val = val.substr(0, len) + "...";
 
-  return "<String: \"" + val + "\">";
+  std::stringstream ss;
+  ss << rang::fg::yellow << "<String: \"" + val + "\">" << rang::fg::reset;
+  return ss.str().c_str();
```
mmarchini: ditto

src/llv8.cc
```diff
@@ -1031,11 +1095,20 @@ std::string FixedArray::Inspect(InspectOptions* options, Error& err) {
...
+    }
+    res += ">";
+  } else {
+    std::stringstream ss;
+    ss << rang::fg::yellow << res + ">" << rang::fg::reset;
+    res = ss.str();
   }
-
-  return res + ">";
+  return res.c_str();
```
mmarchini: ditto

src/llv8.cc
```diff
@@ -1050,7 +1123,11 @@ std::string FixedArray::InspectContents(int length, Error& err) {
     if (!res.empty()) res += ",\n";
 
     char tmp[128];
-    snprintf(tmp, sizeof(tmp), "    [%d]=", i);
+
+    std::stringstream ss;
+    ss << rang::style::bold << rang::fg::yellow << "    [" << i << "]"
+       << rang::fg::reset << rang::style::reset << "=";
+    snprintf(tmp, sizeof(tmp), "%s", ss.str().c_str());
```
mmarchini: `res += ss.str() + value.Inspect(&options, err);` instead of `snprintf`?

src/llv8.cc
```diff
@@ -1295,13 +1424,25 @@ std::string Map::Inspect(InspectOptions* options, Error& err) {
...
            static_cast<int>(in_object_properties_or_constructor_index),
            static_cast<int>(instance_size), descriptors_obj.raw());
+
+  // snprintf(tmp, sizeof(tmp),
+  //          "<Map own_descriptors=%d %s=%d instance_size=%d "
+  //          "descriptors=0x%016" PRIx64,
+  //          static_cast<int>(own_descriptors_count),
+  //          in_object_properties_or_constructor.c_str(),
+  //          static_cast<int>(in_object_properties_or_constructor_index),
+  //          static_cast<int>(instance_size), descriptors_obj.raw());
```
mmarchini: Please remove commented code.

src/llv8.cc
```diff
@@ -1475,7 +1647,12 @@ std::string JSObject::InspectElements(int64_t length, Error& err) {
     if (!res.empty()) res += ",\n";
 
     char tmp[64];
-    snprintf(tmp, sizeof(tmp), "    [%d]=", static_cast<int>(i));
+
+    std::stringstream ss;
+    ss << rang::style::bold << rang::fg::yellow << "    ["
+       << static_cast<int>(i) << "]" << rang::fg::reset << rang::style::reset
+       << "=";
+    snprintf(tmp, sizeof(tmp), "%s", ss.str().c_str());
```
mmarchini: `res += ss.str() + value.Inspect(&options, err);` instead of `snprintf`?

src/llv8.cc
```diff
@@ -1569,6 +1757,17 @@ std::string JSObject::InspectDescriptors(Map map, Error& err) {
       continue;
     }
 
+    // Skip non-fields for now
```
mmarchini: Remove this code as per https://github.com/nodejs/llnode/pull/221/files#diff-86a9160147028a03800f8ba4fa555285L1530

Drieger:
@mmarchini I think all requests were addressed on last commit

src/llscan.h
```diff
@@ -89,6 +89,9 @@ class FindReferencesCmd : public CommandBase {
     virtual void PrintRefs(lldb::SBCommandReturnObject& result, v8::String& str,
                            Error& err) {}
 
+    std::string GetPropertyReferenceString();
+    std::string GetArrayReferenceString();
+
     static const char* const property_reference_template;
     static const char* const array_reference_template;
```
mmarchini: These two properties can be removed now.

src/llv8.cc
```diff
@@ -1577,6 +1748,16 @@ std::string JSObject::InspectDescriptors(Map map, Error& err) {
       continue;
     }
 
+    if (!descriptors.IsFieldDetails(details)) {
```
mmarchini: Please remove this (lines 1751 to 1760). This was removed by https://github.com/nodejs/llnode/pull/221 to unbreak our tests and probably was reinserted here during rebase.

Honestly, I'm surprised tests are passing even though this code is still here.

mmarchini:
@joyeecheung PTAL. I guess it's ok to introduce [rang](https://github.com/agauniyal/rang) as a dependency, right?

Drieger:
@mmarchini Just addressed last two comments.

joyeecheung:
@mmarchini Won't be able to take a look at this until the weekend, but having a in-tree dependency in the public domain doesn't seem to a problem from me.

Drieger:
@mmarchini Conflicts solved

mmarchini:
Landed in 7b09c15, thank you!

PR #227, opened 2018-09-04
src: show context objects within `findrefs`

Drieger:
When using `findrefs` we should be able to get all references
for the given value, this includes `Context` objects.

Refs: https://github.com/nodejs/llnode/issues/195

Sample output:
screenshot from 2018-09-04 16-10-51

### DIFF
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -612,6 +612,10 @@ void FindReferencesCmd::PrintReferences(SBCommandReturnObject& result,
       //    "\n", type, addr);
     }
   }
+
+  // Print references found directly inside Context objects
+  Error err;
+  scanner->PrintContextRefs(result, err);
 }
 
 
@@ -668,6 +672,45 @@ char** FindReferencesCmd::ParseScanOptions(char** cmd, ScanType* type) {
   return &cmd[optind - 1];
 }
 
+// Walk all contexts previously stored and print search_value_
+// reference if it exists. Not all values are associated with
+// a context object. It seems that Function-Local variables are
+// stored in the stack, and when some nested closure references
+// it is allocated in a Context object.
+void FindReferencesCmd::ReferenceScanner::PrintContextRefs(
+    SBCommandReturnObject& result, Error& err) {
+  ContextVector* contexts = llscan_->GetContexts();
+  v8::LLV8* v8 = llscan_->v8();
+
+  for (auto ctx : *contexts) {
+    Error err;
+    v8::HeapObject context_obj(v8, ctx);
+    v8::Context c(context_obj);
+    v8::Context::Locals locals(&c);
+
+    v8::HeapObject scope_obj(locals.ctx->GetScopeInfo(err));
+    if (err.Fail()) return;
+
+    v8::ScopeInfo scope(scope_obj);
+    for (v8::Context::Locals::Iterator it = locals.begin(); it != locals.end();
+         it++) {
+      if (err.Fail()) {
+        continue;
+      }
+      if ((*it).raw() == search_value_.raw()) {
+        v8::String _name = scope.ContextLocalName(
+            it.current, locals.param_count, locals.stack_count, err);
+        if (err.Fail()) return;
+
+        std::string name = _name.ToString(err);
+        if (err.Fail()) return;
+
+        result.Printf("0x%" PRIx64 ": Context.%s=0x%" PRIx64 "\n", c.raw(),
+                      name.c_str(), search_value_.raw());
+      }
+    }
+  }
+}
 
 void FindReferencesCmd::ReferenceScanner::PrintRefs(
     SBCommandReturnObject& result, v8::JSObject& js_obj, Error& err) {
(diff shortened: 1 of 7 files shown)


src/llscan.cc
```diff
@@ -668,6 +672,59 @@ char** FindReferencesCmd::ParseScanOptions(char** cmd, ScanType* type) {
...
+// it is allocated in a Context object.
+void FindReferencesCmd::ReferenceScanner::PrintContextRefs(
+    SBCommandReturnObject& result, Error& err) {
+  ContextVector* contexts = llscan_->GetContexts();
+  v8::LLV8* v8 = llscan_->v8();
+
+  ContextVector visited_contexts;
+  for (auto ctx : *contexts) {
+    if (std::find(visited_contexts.begin(), visited_contexts.end(), ctx) !=
+        visited_contexts.end())
```
mmarchini: Why do we need to keep track of visited contexts? Can we have duplicates on LLScan::contexts_?

src/llscan.cc
```diff
@@ -668,6 +672,59 @@ char** FindReferencesCmd::ParseScanOptions(char** cmd, ScanType* type) {
...
+    if (err.Fail()) return;
+    v8::Smi stack_count_smi = scope.StackLocalCount(err);
+    if (err.Fail()) return;
+    v8::Smi local_count_smi = scope.ContextLocalCount(err);
+    if (err.Fail()) return;
+
+    int param_count = param_count_smi.GetValue();
+    int stack_count = stack_count_smi.GetValue();
+    int local_count = local_count_smi.GetValue();
+    for (int i = 0; i < local_count; i++) {
```
mmarchini: I think we should move this logic to an iterator, since this is basically the same thing we have in https://github.com/nodejs/llnode/blob/master/src/llv8.cc#L1091-L1158.

src/llscan.cc
```diff
@@ -668,6 +672,59 @@ char** FindReferencesCmd::ParseScanOptions(char** cmd, ScanType* type) {
...
+      v8::Value value = c.ContextSlot(i, err);
+      if (value.raw() == search_value_.raw()) {
+        v8::String _name =
+            scope.ContextLocalName(i, param_count, stack_count, err);
+        if (err.Fail()) return;
+
+        std::string name = _name.ToString(err);
+        if (err.Fail()) return;
+
+        result.Printf("0x%" PRIx64 ": (context).%s=0x%" PRIx64 "\n", c.raw(),
```
mmarchini: s/(context)/Context/

To keep coherency with `v8 inspect` output.

src/llv8.cc
```diff
@@ -876,6 +876,10 @@ std::string HeapObject::GetTypeName(Error& err) {
   if (type == v8()->types()->kMapType) {
     return "(Map)";
   }
+  if (type >= v8()->types()->kFirstContextType &&
+      type <= v8()->types()->kLastContextType) {
+    return "(Context)";
```
mmarchini: s/(Context)/Context/

To keep coherency with v8 inspect output.

Drieger:
@mmarchini I just finished suggested changes

src/llscan.cc
```diff
@@ -668,6 +672,45 @@ char** FindReferencesCmd::ParseScanOptions(char** cmd, ScanType* type) {
...
+  ContextVector* contexts = llscan_->GetContexts();
+  v8::LLV8* v8 = llscan_->v8();
+
+  for (auto ctx : *contexts) {
+    Error err;
+    v8::HeapObject context_obj(v8, ctx);
+    v8::Context c(context_obj);
+    v8::Context::Locals locals(&c);
+
+    v8::HeapObject scope_obj(locals.ctx->GetScopeInfo(err));
```
mmarchini: nit: could just be `c.GetScopeInfo(err)`

src/llv8.h
```diff
@@ -397,6 +397,35 @@ class Context : public FixedArray {
...
+     public:
+      Value operator*();
+      const Context::Locals::Iterator operator++(int);
+      bool operator!=(Context::Locals::Iterator that);
+
+      inline Iterator(int current, Context* ctx) : current(current), ctx(ctx){};
+
+     public:
+      int current;
+      Context* ctx;
```
mmarchini: Keep these private and use an accessor method.

src/llv8.h
```diff
@@ -397,6 +397,35 @@ class Context : public FixedArray {
...
+    Locals(Context* ctx_);
+
+    Iterator begin();
+    Iterator end();
+
+   public:
+    int local_count;
+    int param_count;
+    int stack_count;
+    Context* ctx;
```
mmarchini: Ditto.

Drieger:
@mmarchini Last three comments were addressed on last commit

test/plugin/scan-test.js
```diff
@@ -65,6 +65,7 @@ function test(executable, core, t) {
     }
     t.ok(found, 'Zlib should be in findjsinstances');
 
+    sess.send("target modules dump symtab");
```
joyeecheung: Can you separate this out into another test case? (That is, put this in the next `sess.linesUntil` callback, call another `sess.send('version');` there, and move the tests below into a new `sess.linesUntil(versionMark...)` session.) That way it would be easier to debug without mixing the `target modules dump symtab` output in the middle.

(also nit: single quotes)

src/llv8.cc
```diff
@@ -1145,19 +1168,20 @@ std::string Context::Inspect(InspectOptions* options, Error& err) {
...
+    String name = it.LocalName(err);
     if (err.Fail()) return std::string();
 
     if (!res.empty()) res += ",\n";
 
     res += options->get_indent_spaces() + name.ToString(err) + "=";
     if (err.Fail()) return std::string();
 
-    Value value = ContextSlot(i, err);
+    // Value value = ContextSlot(it.current, err);
```
joyeecheung: Can you remove this line?

Drieger:
@joyeecheung I think all change requests were addressed

I created a different function for the symbol lookup, this will allow to use it anywhere else if necessary.

joyeecheung:
Landed in 089dcc9cfe38eaf018d97055ec73499c8a0a904b, thanks!

Issue #218, opened 2018-07-28
Inspect the `stack` property on Error objects


mmarchini:
It would be useful if llnode could inspect the stack property of an Error object. Since the stack property is an [accessor](https://github.com/v8/v8/blob/8109d7e1f11f42af08bf5cbe5ca1313f14589a55/src/accessors.cc#L1142-L1201), it might be tricky to make it work though.

misterdjules:
> Since the stack property is an accessor, it might be tricky to make it work though.

The stack property is an accessor, but the actual internal data about the actual stack frames is [stored as a symbol property](https://github.com/v8/v8/blob/8109d7e1f11f42af08bf5cbe5ca1313f14589a55/src/accessors.cc#L1152-L1153), isn't it?

misterdjules:
FWIW, this is the poor-man error stack output that I use sometimes with mdb_v8:

```
[root@some-smartos-host ~]# mdb node_bin/node-v6.14.1-linux-x64/bin/node ~/some-core-file
mdb: warning: librtld_db failed to initialize; shared library information will not be available
> ::load /root/mdb_v8_amd64.so
mdb_v8 version: 1.3.0 (release, from 28f2bfa)
V8 version: 5.1.281.111
Autoconfigured V8 support from target
C++ symbol demangling enabled
> ::jsstack -vn0
native: v8::base::OS::Abort+9
native: v8::internal::Runtime_Throw+0x133
        (1 internal frame elided)
js:     throwError
          file: /apps/myapp/my-js-file.js
          posn: line 40
          this: 240c23429e11 (JSObject: Immediate)
          arg1: 240c23429921 (JSObject: TypeError)
js:     runCallback
          file: timers.js
          posn: line 666
          this: 90c3f604381 (Oddball: "undefined")
          arg1: 240c23429e11 (JSObject: Immediate)
js:     tryOnImmediate
          file: timers.js
          posn: line 641
          this: 90c3f604381 (Oddball: "undefined")
          arg1: 240c23429e11 (JSObject: Immediate)
          arg2: 240c2342f531 (JSObject: Immediate)
js:     processImmediate
          file: timers.js
          posn: line 592
          this: 3705489827b1 (JSObject: process)
        (1 internal frame elided)
        (1 internal frame elided)
native: v8::internal::Execution::Call+0xc4
native: v8::Function::Call+0x139
native: v8::Function::Call+0x41
native: node::MakeCallback+0x1bc
[...]

Issue #196, opened 2018-05-08
Segmentation fault with `v8 findrefs -s` on master with aborted process inside lldb


mmarchini:
* **Version**: llnode master, node v8.11.1
* **Platform**: OS X (10.13.4)
* **Subsystem**: findrefs

Sometimes I'm getting a Segmentation Fault when I try to use `v8 findrefs -s` on master. Weirest thing is most of the times it just work, making this one a little tricky to track down. Creating an issue to remind myself to look into it later.

Example:

**index.js**
```js
class Lero {
  constructor() {
    this.lala = {};
  }
}

class Lira {
  constructor() {
    this.lele = {};
  }
}


const lero = new Lero();

function foo() {
  const lero2 = new Lero();
  boom();
}

foo();
```

**llnode results**
```
$ ./llnode.sh node
(lldb) target create "node"
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "/usr/local/Cellar/python@2/2.7.14_3/Frameworks/Python.framework/Versions/2.7/lib/python2.7/copy.py", line 52, in <module>
    import weakref
  File "/usr/local/Cellar/python@2/2.7.14_3/Frameworks/Python.framework/Versions/2.7/lib/python2.7/weakref.py", line 14, in <module>
    from _weakref import (
ImportError: cannot import name _remove_dead_weakref
Current executable set to 'node' (x86_64).
(lldb) plugin load /Users/mmarchini/workspace/sthima/llnode/llnode.dylib
(lldb) settings set prompt '(llnode) '
(llnode) r --abort-on-uncaught-exception index.js
Process 69150 launched: '/Users/mmarchini/.nvm/versions/node/v8.11.1/bin/node' (x86_64)
Uncaught ReferenceError: boom is not defined

FROM
foo (/Users/mmarchini/workspace/sthima/llnode/index.js:1:1)
Object.<anonymous> (/Users/mmarchini/workspace/sthima/llnode/index.js:1:1)
Module._compile (module.js:1:1)
Object.Module._extensions..js (module.js:1:1)
Module.load (module.js:1:1)
tryModuleLoad (module.js:1:1)
Function.Module._load (module.js:1:1)
Function.Module.runMain (module.js:1:1)
startup (bootstrap_node.js:1:1)
bootstrap_node.js:1:1
Process 69150 stopped
[...]

PR #233, opened 2018-09-20
src: print stack for error objects

mmarchini:
Error objects have a .stack property, which is actually an accessor. The
stack is stored in a unnamed symbol property and converted to a string
when .stack is accessed in JavaScript. This patch implements the
accessor behavior to print the error stack when inspecting an Error
object.

Fixes: https://github.com/nodejs/llnode/issues/218

### DIFF
--- a/src/llv8-inl.h
+++ b/src/llv8-inl.h
@@ -44,7 +44,6 @@ inline bool HeapObject::Check() const {
   return (raw() & v8()->heap_obj()->kTagMask) == v8()->heap_obj()->kTag;
 }
 
-
 int64_t HeapObject::LeaField(int64_t off) const {
   return raw() - v8()->heap_obj()->kTag + off;
 }
@@ -81,6 +80,21 @@ inline int64_t HeapObject::GetType(Error& err) {
 
   Map map(obj);
   return map.GetType(err);
+}
+
+
+inline bool HeapObject::ISJSErrorType(Error& err) {
+  int64_t type = GetType(err);
+  if (type == v8()->types()->kJSErrorType) return true;
+
+  // NOTE (mmarchini): We don't have a JSErroType constant on Node.js v6.x,
+  // thus we try to guess if the object is an Error object by checking if its
+  // name is Error. Should work most of the time.
+  if (!JSObject::IsObjectType(v8(), type)) return false;
+  JSObject obj(this);
+  if (obj.GetTypeName(err) != "Error" || err.Fail()) return false;
+
+  return true;
 }
 
 
--- a/src/llv8-inl.h
+++ b/src/llv8-inl.h
@@ -44,6 +44,7 @@ inline bool HeapObject::Check() const {
   return (raw() & v8()->heap_obj()->kTagMask) == v8()->heap_obj()->kTag;
 }
 
+
 int64_t HeapObject::LeaField(int64_t off) const {
   return raw() - v8()->heap_obj()->kTag + off;
 }
@@ -83,18 +84,19 @@ inline int64_t HeapObject::GetType(Error& err) {
 }
 
 
-inline bool HeapObject::ISJSErrorType(Error& err) {
+inline bool HeapObject::IsJSErrorType(Error& err) {
   int64_t type = GetType(err);
+  if (err.Fail()) return false;
   if (type == v8()->types()->kJSErrorType) return true;
 
-  // NOTE (mmarchini): We don't have a JSErroType constant on Node.js v6.x,
+  // NOTE (mmarchini): We don't have a JSErrorType constant on Node.js v6.x,
   // thus we try to guess if the object is an Error object by checking if its
   // name is Error. Should work most of the time.
   if (!JSObject::IsObjectType(v8(), type)) return false;
-  JSObject obj(this);
-  if (obj.GetTypeName(err) != "Error" || err.Fail()) return false;
 
-  return true;
+  JSObject obj(this);
+  std::string type_name = obj.GetTypeName(err);
+  return err.Success() && type_name == "Error";
 }
 
 
--- a/src/llv8.cc
+++ b/src/llv8.cc
@@ -827,6 +827,11 @@ std::string HeapObject::Inspect(InspectOptions* options, Error& err) {
     return pre + m.Inspect(options, err);
   }
 
+  if (ISJSErrorType(err)) {
+    JSError error(this);
+    return pre + error.Inspect(options, err);
+  }
+
   if (JSObject::IsObjectType(v8(), type)) {
     JSObject o(this);
     return pre + o.Inspect(options, err);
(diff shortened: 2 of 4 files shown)


mmarchini:
Jenkins CI: https://ci.nodejs.org/view/post-mortem/job/llnode-pipeline/31/

(first try, let's see how it goes)

src/llv8-inl.h
```diff
@@ -84,6 +83,21 @@ inline int64_t HeapObject::GetType(Error& err) {
 }
 
 
+inline bool HeapObject::ISJSErrorType(Error& err) {
+  int64_t type = GetType(err);
```
cjihrig: Missing `err.Fail()` check?

src/llv8-inl.h
```diff
@@ -84,6 +83,21 @@ inline int64_t HeapObject::GetType(Error& err) {
...
+  if (type == v8()->types()->kJSErrorType) return true;
+
+  // NOTE (mmarchini): We don't have a JSErroType constant on Node.js v6.x,
+  // thus we try to guess if the object is an Error object by checking if its
+  // name is Error. Should work most of the time.
+  if (!JSObject::IsObjectType(v8(), type)) return false;
+  JSObject obj(this);
+  if (obj.GetTypeName(err) != "Error" || err.Fail()) return false;
```
cjihrig: Can you make the `err.Fail()` check first?

src/llv8-inl.h
```diff
@@ -84,6 +83,21 @@ inline int64_t HeapObject::GetType(Error& err) {
...
+
+  return true;
```
cjihrig: I think this line and the `if` statement above could be combined into a single `return` statement.

src/llv8.cc
```diff
@@ -1523,15 +1537,101 @@ std::string JSObject::Inspect(InspectOptions* options, Error& err) {
...
+    // TODO (mmarchini): once we have Symbol support we'll need to search for
+    // <unnamed symbol>, since the stack symbol doesn't have an external name.
+    // In the future we can add postmortem metadata on V8 regarding existing
+    // symbols, but for now we'll use an heuristic to find the stack in the
+    // error object.
+    Value maybe_stack = GetProperty("<non-string>", err);
+
+    if (err.Fail()) {
+      Error::PrintInDebugMode(
+          "Couldn't find a symbol property in the Error objcet.");
```
cjihrig: objcet -> object

mmarchini:
@cjihrig thank you for your review! I addressed your comments, PTAL.

src/llv8.cc
```diff
@@ -1523,15 +1537,101 @@ std::string JSObject::Inspect(InspectOptions* options, Error& err) {
...
+
+    if (err.Fail()) {
+      Error::PrintInDebugMode(
+          "Couldn't get the first element from the stack array");
+      return res;
+    }
+
+    int64_t stack_len = Smi(maybe_stack_len).GetValue();
+
+    int multiplier = 5;
```
joyeecheung: Maybe it would be more maintainable if `FrameArray::kElementsPerFrame` is exposed in the post mortem metadata and gets loaded here..(not sure if that's on 6.x but it's still in TOT)

src/llv8.cc
```diff
@@ -1523,15 +1537,101 @@ std::string JSObject::Inspect(InspectOptions* options, Error& err) {
...
+
+    std::stringstream error_stack;
+    error_stack << std::endl
+                << rang::fg::red << "  error stack" << rang::fg::reset << " {"
+                << std::endl;
+
+    // TODO (mmarchini): Refactor: create an StackIterator which returns
+    // StackFrame objects
+    for (int64_t i = 0; i < stack_len; i++) {
+      Value maybe_fn = arr.GetArrayElement(2 + (i * multiplier), err);
```
joyeecheung: So this is actually `kFirstIndex + i * kElementsPerFrame + kFunctionOffset` right?

src/llv8.cc
mmarchini: Yes, I have this on my TODO list and will probably get to that at some point.

mmarchini:
Landed in bb0d400

mmarchini:
Yes, I'll prepare a release today, which should be released in the next few days :)

(Want to make sure we have a new release before #246 lands)

misterdjules:
@mmarchini I was inspecting a core dump generated from node 6.x and I was trying to get the actual stack trace for the error, and when trying to get the _full_ stack trace with `v8 inspect -l length-of-stack-array' I constantly get a segmentation fault:

```
$ lldb ~/Downloads/node-v6.10.3-linux-x64/bin/node -c /path/to/core/dump
(lldb) target create "/Users/jgilli/Downloads/node-v6.10.3-linux-x64/bin/node" --core "/path/to/core/dump"
Core file '/path/to/core/dump' (x86_64) was loaded.
(lldb) plugin load /Users/jgilli/dev/llnode/llnode.dylib
(lldb) v8 bt
 * thread #1: tid = 1024, 0x0000000001334fa9 node`v8::base::OS::Abort() + 9, name = 'node', stop reason = signal SIGILL
  * frame #0: 0x0000000001334fa9 node`v8::base::OS::Abort() + 9
    frame #1: 0x0000000000d5241a node`v8::internal::Isolate::Throw(v8::internal::Object*, v8::internal::MessageLocation*) + 490
    frame #2: 0x0000000000eb4e33 node`v8::internal::Runtime_Throw(int, v8::internal::Object**, v8::internal::Isolate*) + 307
    frame #3: 0x000035e3c31092a7 <exit>
    frame #4: 0x000035e3c58ec96b throwError(this=0x2ad297b2fc41:<Object: Immediate>, 0x2ad297b2f569:<Object: TypeError>) 
[...redacted to hide private info...]
(lldb) v8 inspect 0x2ad297b2f569
0x2ad297b2f569:<Object: TypeError properties {
    .stack=0x2ad297b2f981:<Array: length=45>,
    .<non-string>=0x2ad297b2f981:<Array: length=45>,
    .message=0x2ad297b2f541:<String: "Cannot match aga...">,
    .<non-string>=0x2725531043c1:<true>}>
(lldb) v8 inspect 0x2ad297b2f981
0x2ad297b2f981:<Array: length=45 {
[...redacted to hide private info...]
}>
(lldb) v8 inspect -l 44 0x2ad297b2f981
Segmentation fault: 11
$
```

Am I missing something?

mmarchini:
Seems like a separate issue. `v8 inspect -l 44 0x2ad297b2f981` is inspecting an Array unaware it might be a stack trace, so it's not - or at least shouldn't be - related to changes on this PR.

This is not the first segfault I see in llnode (#196), we probably have some fragile code paths somewhere. Do you mind opening a new issue for this? Also, it would be nice to get llnode's stack trace right before it segfaulted. I haven't tried it, but it seems to be possible with dtrace: https://stackoverflow.com/a/17115382/2956796.

misterdjules:
@mmarchini How are users supposed to inspect the stack trace? `v8 inspect address-of-stack-property` doesn't seem to output the full stack.

mmarchini:
`v8 inspect 0x2ad297b2f569` (the Error object) should give you the stack trace.

Ohhh I see, this is a `TypeError`. Ok, we'll need to [handle this differently in Node.js v6.x](https://github.com/nodejs/llnode/blob/master/src/llv8-inl.h#L92-L99).
Will do a follow-up PR for that.

mmarchini:
FWIW, it's working as expected on Node.js v10.13.0

```
(llnode) v8 findjsinstances -d TypeError
0x17febe244e11:<Object: TypeError properties {
    .stack=0x17feba59f0d9:<unknown>,
    .<non-string>=0x17febe244fd1:<Array: length=41>}
  error stack {
    0x17fe6794dd81:<function: (anonymous) at [eval]:1:0>
    0x17febe210d61:<function: runInThisContext at vm.js:91:19>
    0x17febe210c71:<function: runInThisContext at vm.js:299:26>
    0x17febe244791:<function: (anonymous) at [eval]-wrapper:1:10>
    0x17fe27675eb9:<function: Module._compile at (external).js:649:37>
    0x17fe6791fae9:<function: evalScript at internal/bootstrap/node.js:1:10>
    0x17fec0b985a9:<function: startup at internal/bootstrap/node.js:1:10>
    0x17fec0b987e1:<function: bootstrapNodeJSCore at internal/bootstrap/node.js:1:10>
  }>
(Showing 1 to 1 of 1 instances)
```

joyeecheung:
Tips: you can turn on the environment variable `LLNODE_DEBUG=true` to see some debug output from llnode. Also running `log enable lldb api` in the lldb console would tell lldb to log all the API calls which may help as well (if you don't want to debug into the plugin, that'll be debugging the debugger)

PR #235, opened 2018-10-01
Show a limited number of entries

Drieger:
Added a new option to `findjsinstances` that allow user to limit the number of entries shown. Command would be as follow
```
v8 findjsinstances -n <num> <search>
```
First `num` entries will be displayed, to see the following `num` instances user would just repeat the same `num` and `search`, or just press [ENTER] as it repeats the last command. Once we reached the last "page" it begins to show from the first entry again.

I created a simple example to show usage:
https://asciinema.org/a/X04QMdeyWlDY1DBWeCDBNVUry

### DIFF
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -35,7 +35,6 @@ const char* const
 const char* const FindReferencesCmd::ObjectScanner::array_reference_template =
     "0x%" PRIx64 ": %s[%" PRId64 "]=0x%" PRIx64 "\n";
 
-
 const char* const
     FindReferencesCmd::StringScanner::property_reference_template =
         "0x%" PRIx64 ": %s.%s=0x%" PRIx64 " '%s'\n";
@@ -283,7 +282,9 @@ bool FindInstancesCmd::DoExecute(SBDebugger d, char** cmd,
       std::string res = v8_value.Inspect(&inspect_options, err);
       result.Printf("%s\n", res.c_str());
     }
-    result.Printf("..........\n");
+    if (it != t->GetInstances().end()) {
+      result.Printf("..........\n");
+    }
     result.Printf("(Showing %d to %d of %d instances)\n", initial_p_offset + 1,
                   final_p_offset, pagination_.total_entries);
 
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -7,6 +7,7 @@
 #include <cinttypes>
 #include <fstream>
 #include <vector>
+#include <iostream>
 
 #include <lldb/API/SBExpressionOptions.h>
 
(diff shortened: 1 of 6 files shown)


src/llscan.h
```diff
@@ -18,6 +18,19 @@ typedef std::map<uint64_t, ReferencesVector*> ReferencesByValueMap;
 typedef std::map<std::string, ReferencesVector*> ReferencesByPropertyMap;
 typedef std::map<std::string, ReferencesVector*> ReferencesByStringMap;
 
+
+// New type defining pagination options
+// It should be feasible to use it to any commands that output
+// a list of information
+typedef struct cmd_pagination {
```
mmarchini: 1. I'd rather have this as a class, but we can keep this as a struct.
2. `typedef` is not required for structs in C++ (C++ structs are not the same as C structs).

src/llscan.cc
```diff
@@ -230,15 +236,53 @@ bool FindInstancesCmd::DoExecute(SBDebugger d, char** cmd,
...
+            : std::next(t->GetInstances().begin(), initial_p_offset);
+    for (; it != t->GetInstances().end() &&
+           it != (std::next(t->GetInstances().begin(), final_p_offset));
+         ++it) {
       Error err;
       v8::Value v8_value(llscan_->v8(), *it);
       std::string res = v8_value.Inspect(&inspect_options, err);
       result.Printf("%s\n", res.c_str());
     }
+    result.Printf("..........\n");
```
mmarchini: Suggestion: Only print `"..........\n"` if there are more objects to be printed.

src/llscan.cc
```diff
@@ -230,15 +236,53 @@ bool FindInstancesCmd::DoExecute(SBDebugger d, char** cmd,
...
       llscan_->GetMapsToInstances().find(type_name);
+
   if (instance_it != llscan_->GetMapsToInstances().end()) {
     TypeRecord* t = instance_it->second;
-    for (std::set<uint64_t>::iterator it = t->GetInstances().begin();
-         it != t->GetInstances().end(); ++it) {
+
+    // Update pagination options
+    if (full_cmd != pagination_.command ||
+        inspect_options.output_limit != pagination_.output_limit) {
```
mmarchini: Doesn't `full_cmd != pagination_.command` imply `inspect_options.output_limit != pagination_.output_limit`? Do we really need both?

Drieger: @mmarchini `full_cmd` is only the serch term here. Thus we need check if limit is the same.

mmarchini:
Landed in 2fc1571. Thank you!

Issue #238, opened 2018-10-04
Significant performance overhead on 2.0.0


mmarchini:
I'm trying to `v8 findjsobjects` on a ~1.6GB core dump. On v1.7.1 the command finishes after 1 minute whereas on v2.0.0 it's taking more than 10 minutes (couldn't finish one run yet). I'll try to provide a reproducible test case later.

mmarchini:
This regression was introduced in 089dcc9 (which makes sense since my code is OOMing because of a recursive function). Hopefully, there's a way to optimize this.

PR #239, opened 2018-10-04
src: fix performance regression

mmarchini:
Performance overhead was huge when we had a huge amount of contexts.
Two improvements were necessary to reduce the overhead:

  - Keep a cache of context maps (just like we do for object maps)
  - Use std::unordered_set instead of std::set to store contexts
    - Also use it on TypeRecord::instances_, since it is faster
    - Don't use find before insert, sets don't have duplicates by
      definition

Fixes: https://github.com/nodejs/llnode/issues/238


mmarchini:
cc @cjihrig @joyeecheung @Drieger

mmarchini:
Landed in 388605f

Issue #115, opened 2017-07-10
v8 findrefstree command


yunong:
It'd be really useful to have a `v8 findrefstree` command for solving memory leaks. This would take a reference to an object, and starting from that node, print out all objects that refer to that object, and its parents, recursively. 

Generally speaking this would be helpful for debugging memory leaks, as it's usually pretty obvious what objects are leaking, but not the root object(s) that's holding on to the leaking objects. Currently you'd have to run `v8 findrefs -r` over and over an object and its parent until you reached the root, which is slow and very tedious.

Having a command that could recursively print out the tree of objects that refer to a specific object would simplify the debugging process for memory leaks.

hhellyer:
I think we could probably do this but before diving into the technical side what would the output look like?
We need to build a path from an object (through many other objects) to possibly a number of roots. I'm not quite sure how to present that in a terminal window in a way that's usable. I've seen tools for walking the incoming references to an object before but they've generally been in GUIs.

The resulting tree can be very deep (if it passes through structures like linked lists) and very wide (if it ever passes through objects with many incoming references). I'm not sure how to make it presentable in a way that's useful.

yunong:
@hhellyer perhaps we can generate it into a common format that we've been discussing over here? https://github.com/nodejs/post-mortem/issues/13

lldb may not be the best tool to search through the generated tree, but if it's exported in a common format we could write tools to import it into a database and run queries.

Thoughts?

hhellyer:
I'm not sure if we could export the heap from a core dump into another format correctly. We currently have no way to get all the heap roots out of a core dump. (I'm fairly sure I've discussed this somewhere else but I can't find the reference.)

Getting the incoming references is something you need as soon as you start doing any kind of heap dump analysis so it would be worth thinking about and discussing on that thread
 It might not be something we'd want to put into the dump as it's generated but you could add them via post processing or simply generate a dump in the same format with the references reversed. There might need to be a header flag saying which way the references went or you might use the two dumps as a pair - this all probably needs more thought but that should probably happen over in the discussions of the common format.

rnchamberlain:
IMO the best way to meet this requirement would be a GUI front end to the core-dump reader, so you can click on an object and open up a 'reversed' view of the objects that refer to it, with collapsed/expanded views of arrays etc. We are some way from having that. However, I think we could do something in line-mode if we control the amount of output using counts and ".... 200 more like this" type messages, with options to expand fully.

mmarchini:
[...]

PR #244, opened 2018-10-23
src: add recursive option to `findrefs` command

Drieger:
Introduce a new command flag to `findrefs`, command --recursive
allow user to transverse the whole references tree for the given
object.

Refs: https://github.com/nodejs/llnode/issues/115


src/llscan.cc
```diff
@@ -645,11 +653,47 @@ void FindReferencesCmd::PrintReferences(SBCommandReturnObject& result,
...
-      scanner->PrintRefs(result, js_obj, err);
+      scanner->PrintRefs(result, js_obj, err, level);
+
+      if (options->recursive_scan) {
+        if (find(already_visited_references->begin(), already_visited_references->end(), addr) != already_visited_references->end()) {
+          Settings* settings = Settings::GetSettings();
+          unsigned int padding = settings->GetTreePadding();
+          std::stringstream seen_str;
+          seen_str << std::string((level) * padding, ' ') << "|\n";
+          seen_str << std::string((level) * padding, ' ') << "+" << std::string(padding, '-') << rang::fg::red << " [seen above]" << rang::fg::reset << std::endl;
```
mmarchini: maybe a softer color, like gray?

src/llscan.cc
```diff
@@ -645,11 +653,47 @@ void FindReferencesCmd::PrintReferences(SBCommandReturnObject& result,
...
+          result.Printf(seen_str.str().c_str());
+        } else {
+          already_visited_references->push_back(addr);
+          v8::Value value(llscan_->v8(), addr);
+          ObjectScanner* scanner_ = new ReferenceScanner(llscan_, value);
+          ReferencesVector* references_ = scanner_->GetReferences();
```
mmarchini: Can't we use `references` instead `scanner_->GetReferences()`?

src/llscan.cc
```diff
@@ -645,11 +653,47 @@ void FindReferencesCmd::PrintReferences(SBCommandReturnObject& result,
...
+          Settings* settings = Settings::GetSettings();
+          unsigned int padding = settings->GetTreePadding();
+          std::stringstream seen_str;
+          seen_str << std::string((level) * padding, ' ') << "|\n";
+          seen_str << std::string((level) * padding, ' ') << "+" << std::string(padding, '-') << rang::fg::red << " [seen above]" << rang::fg::reset << std::endl;
+          result.Printf(seen_str.str().c_str());
+        } else {
+          already_visited_references->push_back(addr);
+          v8::Value value(llscan_->v8(), addr);
+          ObjectScanner* scanner_ = new ReferenceScanner(llscan_, value);
```
mmarchini: `ObjectScanner scanner_(llscan_, value);`, so we can drop the delete below and let C++ manage memory for us :)

src/llscan.cc
```diff
@@ -645,11 +653,47 @@ void FindReferencesCmd::PrintReferences(SBCommandReturnObject& result,
...
+          delete scanner_;
+        }
+      }
 
     } else if (type < v8->types()->kFirstNonstringType) {
       v8::String str(heap_object);
-      scanner->PrintRefs(result, str, err);
+      scanner->PrintRefs(result, str, err, level);
+
+      if (options->recursive_scan) {
```
mmarchini: Same comments as above

src/llscan.cc
```diff
@@ -746,24 +795,68 @@ void FindReferencesCmd::ReferenceScanner::PrintContextRefs(
...
+          ss << std::string((level-1) * padding, ' ') << "+" << std::string(padding-1, '-') << "+ ";
+        }
+        ss << rang::fg::cyan << "0x%" PRIx64  << rang::fg::reset << ": "
+           << rang::fg::magenta << "Context" << rang::style::bold << rang::fg::yellow
+           << ".%s" << rang::fg::reset << rang::style::reset << "=" << rang::fg::cyan
+           << "0x%" PRIx64 << rang::fg::reset << "\n";
+
+        result.Printf(ss.str().c_str(), c.raw(), name.c_str(), search_value_.raw());
+
+        if (options->recursive_scan) {
```
mmarchini: Same comments as above

src/llscan.cc
```diff
@@ -746,24 +795,68 @@ void FindReferencesCmd::ReferenceScanner::PrintContextRefs(
...
+        if (level > 0) {
+          Settings* settings = Settings::GetSettings();
+          unsigned int padding = settings->GetTreePadding();
+          ss << std::string((level-1) * padding, ' ') << "|\n";
+          ss << std::string((level-1) * padding, ' ') << "+" << std::string(padding-1, '-') << "+ ";
+        }
+        ss << rang::fg::cyan << "0x%" PRIx64  << rang::fg::reset << ": "
+           << rang::fg::magenta << "Context" << rang::style::bold << rang::fg::yellow
+           << ".%s" << rang::fg::reset << rang::style::reset << "=" << rang::fg::cyan
+           << "0x%" PRIx64 << rang::fg::reset << "\n";
```
mmarchini: Can't we call `Context::Inspect` instead of reformatting it here?

src/llnode.cc
```diff
@@ -135,6 +135,29 @@ bool SetPropertyColorCmd::DoExecute(SBDebugger d, char** cmd,
...
+
+  if (cmd == nullptr || *cmd == nullptr) {
+    result.SetError("USAGE: v8 settings set tree-padding [1..10]");
+    return false;
+  }
+  Settings* settings = Settings::GetSettings();
+  std::stringstream option(cmd[0]);
+  int padding;
+
+  if (!(option >> padding)) {
```
mmarchini: Add a comment explaining this converts string to int and returns false if it can't.

Drieger: Added an explanation

src/llscan.cc
Drieger: This is a new reference vector, base on the new `scanner_`. If we use the already existent `references` we would overwrite and maybe lose some values.

Drieger:
@mmarchini I already addressed your comments.

src/llscan.cc
```diff
@@ -645,11 +650,55 @@ void FindReferencesCmd::PrintReferences(SBCommandReturnObject& result,
       // Basically we need to access objects and arrays as both objects and
       // arrays.
       v8::JSObject js_obj(heap_object);
-      scanner->PrintRefs(result, js_obj, err);
+      scanner->PrintRefs(result, js_obj, err, level);
+
+      if (options->recursive_scan) {
```
mmarchini: This if block seems identical to lines 682-699 and 824-843. Can we reuse this somehow?

src/llnode.cc
```diff
@@ -491,6 +522,7 @@ bool PluginInitialize(SBDebugger d) {
       " * -n, --name  name     - all properties with the specified name\n"
       " * -s, --string string  - all properties that refer to the specified "
       "JavaScript string value\n"
+      " * -r, --recursive      - walk through references tree recursively\n"
```
mmarchini: `-r` is incompatible with `-s` and `-n`, right? If so this should be documented, otherwise, we should have tests for `-r -s <string>` and `-r -n <property>`.

Drieger:
@mmarchini I moved duplicated code to a method and added tests for all `findrefs` options.

Drieger:
The error seems to be related to https://github.com/nodejs/llnode/issues/267 and was not introduced by this PR, it is being captured by the test in some situations.

mmarchini:
This is the result for a recursive findrefs on Zlib:

image

This feature will be really helpful 😄

mmarchini:
I'll keep this open for a few days in case someone else wants to review it, otherwise I plan to land this by the weekend.

mmarchini:
Landed in 85f067a...c29d5af, thank you!

src/llnode.h
```diff
@@ -30,6 +30,18 @@ class SetPropertyColorCmd : public CommandBase {
...
+class SetTreePaddingCmd : public CommandBase {
+ public:
+  SetTreePaddingCmd(v8::LLV8* llv8) : llv8_(llv8) {}
+  ~SetTreePaddingCmd() override {}
+
+  bool DoExecute(lldb::SBDebugger d, char** cmd,
+                 lldb::SBCommandReturnObject& result) override;
+
+ private:
+  v8::LLV8* llv8_;
```
fanatid: `llv8_` unused, PR with fix #297

Issue #245, opened 2018-10-25
llnode warnings during compilation process


Drieger:
Current compilation process throws some warnings. Wouldn't it be a good approach to prevent merge of PRs that introduce new warnings? Or maybe add it to the test pipeline?

Current warnings are as follow:
```
llv8.h:395:17: warning: inline function ‘llnode::v8::String llnode::v8::ScopeInfo::ContextLocalName(int, int, int, llnode::Error&)’ used
 but never defined
```

```
llv8.h:410:16: warning: inline function ‘llnode::v8::Value llnode::v8::Context::ContextSlot(int, llnode::Error&)’ used but never defined
```

```
llv8.cc:107:43: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
   return *reinterpret_cast<double*>(&value);
```

```
llv8.cc:1969:14: warning: variable ‘value’ set but not used [-Wunused-but-set-variable]
```

```
llscan.cc:1468:8: warning: variable ‘done’ set but not used [-Wunused-but-set-variable]
```

```
llscan.cc:528:36: warning: ‘scanner’ may be used uninitialized in this function [-Wmaybe-uninitialized]
```

```

PR #246, opened 2018-11-01
src: introduce Inspector class

mmarchini:
Inspect methods (which are methods to visualize content from an object)
were tightly coupled with object definitions. In fact, the structure of
some objects is defined inside those Inspect methods, making it harder
to reuse code (for example, to return structured objects in our API
instead of strings). This commit is the beginning of a refactor effort
to decouple object visualization and object strucutres in order to
provide simpler code and a more useful API.

Most changes in this commit are moving ::Inspect methods from V8 Object
classes to a separate class Inspector, and changing call sites to use
this new approach. Inspector::Inspect is a single templated method with
specialization for each class it knows how to display.


mmarchini:
By the way, I might have fixed some warning as well (#245).

mmarchini:
cc @joyeecheung @indutny @cjihrig @bnoordhuis @rnchamberlain @hhellyer @richardlau @Drieger since I don't want this to land unnoticed.

joyeecheung:
Great work! From a glance the approach looks reasonable to me, two things to note:

- The name inspector sort of gave me a false impression that this is implementing the inspector protocol..maybe viewer would be a less ambiguous name, but I don't feel particularly strongly about this
- The linux tests seem to be failing on Travis?

mmarchini:
Renamed `Inspector` to `Printer`, `Inspect*` to `Stringify*` and `InspectOptions` to `PrinterOptions`. I'll try to look into the Linux failures next.

mmarchini:
Seems to be a linking problem:

```
(llnode) v8 inspect 0x255469f99d69
/usr/local/bin/lldb: symbol lookup error: /home/mmarchini/llnode/llnode.so: undefined symbol: _ZN6llnode2v88JSObject16GetInObjectValueIdEET_liRNS_5ErrorE
$ echo "_ZN6llnode2v88JSObject16GetInObjectValueIdEET_liRNS_5ErrorE" | c++filt
double llnode::v8::JSObject::GetInObjectValue<double>(long, int, llnode::Error&)
```

mmarchini:
Fixed by moving `JSObject::GetInObjectValue` implementation to `src/llv8-inl.h`.

mmarchini:
Rebased, but I want to make a release before landing this (just in case this breaks something).

mmarchini:
Landed in c851b9a

mmarchini:
I missed the failing CI on Linux before landing 😕

Will open a PR to unbreak our CI soon.

Issue #232, opened 2018-09-17
Upcoming metadata changes in V8 7.0


cjihrig:
- `v8dbg_class_SharedFunctionInfo__script__Object`
  Use: `v8dbg_class_SharedFunctionInfo__script_or_debug_info__Object`

- `v8dbg_class_SharedFunctionInfo__function_identifier_or_debug_info__Object`
  Use: `v8dbg_class_UncompiledData__inferred_name__String` and
           `v8dbg_class_SharedFunctionInfo__name_or_scope_info__Object`
    
Refs: https://github.com/v8/v8/commit/c941f11abddf57844e9807503df1877f6f60c94a

mmarchini:
Fixed by #247

Issue #226, opened 2018-09-02
Upcoming metadata changes in V8 6.9


cjihrig:
The V8 6.9 update requires the following adjustments to the
postmortem debugging metadata constants:

- `v8dbg_class_SharedFunctionInfo__function_identifier__Object`
  Use: `v8dbg_class_SharedFunctionInfo__function_identifier_or_debug_info__Object`

- `v8dbg_class_SharedFunctionInfo__end_position__int`
  Use: `v8dbg_class_UncompiledData__end_position__int32_t`

- `v8dbg_class_SharedFunctionInfo__start_position_and_type__int`
  Use: `v8dbg_class_UncompiledData__start_position__int32_t`

- `v8dbg_sharedfunctioninfo_start_position_mask` and `v8dbg_sharedfunctioninfo_start_position_shift`
  Removed. Use: `v8dbg_class_UncompiledData__start_position__int32_t`

- `v8dbg_scopeinfo_idx_nstacklocals` has been removed

See: https://github.com/nodejs/node/pull/21983/commits/541e8c2da17146a731891d880841169ac2c60a06
Refs: https://github.com/v8/v8/commit/39e2d97bc4d6b35be40294f06873f03b0368ce5c
Refs: https://github.com/v8/v8/commit/467eb1497e998dd73a4006cf0655fd0ca289b6f4
Refs: https://github.com/v8/v8/commit/c51bcd17bb4da87e1b9bf067351f0fc5cce4bbd5

mmarchini:
Related: https://github.com/nodejs/llnode/issues/202

mmarchini:
Fixed by #247

Issue #202, opened 2018-06-05
Upcoming metadata changes in V8 6.9


mmarchini:
- v8dbg_class_SharedFunctionInfo__internal_formal_parameter_count__int
  - Renamed: v8dbg_class_SharedFunctionInfo__internal_formal_parameter_count__uint16_t
  - V8 commit: https://github.com/v8/v8/commit/53d4dfc3771dda011ba95d0f825fabb83219ff51
- v8dbg_scopeinfo_idx_nstacklocals
  - Removed
  - V8 commit: v8/v8@467eb14

mmarchini:
Fixed by #247

PR #247, opened 2018-11-06
WIP src: fixes for V8 6.9 and 7.0

mmarchini:
Several fixes for V8 6.9 (#202, #226) and 7.0 (#232). Still a work in progress though, I'm sending a PR to run CI on these changes.

#### TODO

  - [ ] Improve commit message
  - [ ] Check if there's no stalled code in the PR
  - [ ] Add more tests (I think there's one missing frame in the stack trace)


mmarchini:
`Timer` check was removed from workqueue test because of https://github.com/nodejs/node/issues/25806. The test was a duplicate of the previous test anyway, so this shouldn't decrease coverage.

mmarchini:
While working on this I noticed we use `Error::Failures` a lot when trying to decode an object from the heap. The problem is those `Error::Failures` propagate and usually abort our command. 

For example: llnode was unable to output the content of `process` objects in Node.js 11 because it couldn't get the script and source position for some methods in the object, but it could get everything else. llnode didn't give any output to the user because it couldn't retrieve a small piece of information.

IMO we should change the code to be more tolerant when it can't decode only a piece of information. Most users of llnode use it to analyze core dumps, which may have corrupted memory addresses, and being fault-tolerant is essential to analyze such core dumps. 

We should move most occurrences of `Error::Failure` to `Error::PrintInDebugMode` and save `Error::Failures` for unrecoverable errors.

This would also allow llnode to work better for future Node.js versions, with fewer features affected by V8 changes.

mmarchini:
This is ready for review :)

cc @nodejs/llnode

mmarchini:
CI failures are coming from recent changes in core (bisected and found https://github.com/nodejs/node/pull/26382 to be the reason). Since apparently we were relying on a type indirectly defined internally on core (`(Object)`, which is an object with a non-js-function constructor according to llnode code), I dropped this type from our tests. Remaining basic types should still be enough for testing.

mmarchini:
CI is green. I'll start working on fixes for Node.js v12 (V8 7.1 to 7.6) in the next few days in an attempt to have llnode working on 12 soon after it goes LTS. But it would be nice to merge this first. cc @nodejs/llnode @nodejs/diagnostics

PR #252, opened 2018-12-10
src: fix build on older compilers (gcc 4.8)

mmarchini:
Build broke on older compilers (such as gcc 4.8) after
https://github.com/nodejs/llnode/pull/246 landed. The issue is related
to using default values for templates in the ::Stringiy definition. This
commit makes minimal changes to fix the build while keeping
functionality intact. A follow-up refactor of Stringify<JSError> is
required.


mmarchini:
cc @joyeecheung @cjihrig

mmarchini:
Landed in c17d0d1

PR #256, opened 2019-01-16
src: refactor JSError stringifier

mmarchini:
Essentially move all the logic related to how JSError stack traces are
accessible to llv8 files, breaking this logic into smaller functions.
Also moved the chunk of code related to printing JSObject properties to
its own function, to allow reuse between JSError and JSObject
stringifiers.


mmarchini:
Landed in a295239

PR #3, opened 2015-11-09
llnode: add support for JSDate

evanlucas:
Result:

```
(lldb) v8 i 0x000001a3f3124551
0x000001a3f3124551:<JSDate: 1447105503556>

$ node
> new Date(1447105503556)
Mon Nov 09 2015 15:45:03 GMT-0600 (CST)
```

PR #257, opened 2019-01-16
src: move JSDate "ToString" logic to JSDate class

Drieger:
Move logic that was representing a JSDate object
to a string from the Printer::Stringfy method to
the JSDate class, implementing "ToString" method.

### DIFF
--- a/src/llv8-inl.h
+++ b/src/llv8-inl.h
@@ -279,6 +279,18 @@ ACCESSOR(JSRegExp, GetSource, js_regexp()->kSourceOffset, String)
 
 ACCESSOR(JSDate, GetValue, js_date()->kValueOffset, Value)
 
+bool JSDate::IsSmi(Error& err) {
+  v8::Value val(GetValue(err));
+  v8::Smi smi(val);
+  return smi.Check();
+}
+
+bool JSDate::IsHeapNumber(Error& err) {
+  v8::Value val(GetValue(err));
+  v8::HeapNumber heap_number(val);
+  return heap_number.Check();
+}
+
 bool String::IsString(LLV8* v8, HeapObject heap_object, Error& err) {
   if (!heap_object.Check()) return false;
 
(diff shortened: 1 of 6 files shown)


src/printer.cc
```diff
@@ -136,34 +137,10 @@ std::string Printer::Stringify(v8::JSFunction js_function, Error& err) {
...
-  v8::HeapNumber hn(val);
-  if (hn.Check()) {
-    std::string s = hn.ToString(true, err);
-    if (err.Fail()) {
-      return pre + ">";
-    }
-    return pre + s + ">";
-  }
-
-  double d = static_cast<double>(val.raw());
```
Drieger: @evanlucas do you remember why this last part, casting to `double` was added? I couldn't find any test case where I could reach it. It was originally introduced in https://github.com/nodejs/llnode/pull/3/files

evanlucas: not off the top of my head, sorry!

src/llv8-inl.h
```diff
@@ -279,6 +279,18 @@ ACCESSOR(JSRegExp, GetSource, js_regexp()->kSourceOffset, String)
 
 ACCESSOR(JSDate, GetValue, js_date()->kValueOffset, Value)
 
+bool JSDate::IsSmi(Error& err) {
```
joyeecheung: These two helpers seem unnecessary if the call site need to grab the values out either ways?

mmarchini: I like the idea of having IsTYPE methods instead of duplicating the tests everywhere. If V8 changes how we check for a type we only need to make changes to one place. But I'd rather have all IsTYPE methods in the Value class, so it resembles the V8 API (https://github.com/nodejs/llnode/pull/247/files#diff-6863f2f9f81092029c00ed6e07b78700R60).

mmarchini: But looking again those two don't add much value, since they're just calling `.Check`...

We might want to settle on style here (`TYPE.Check()` vs. `Value.IsTYPE()`), but we don't have to decide in this PR.

mmarchini:
Ignore my previous comment, I've mistaken ToString with the old `Inspect` methods. This looks fine.

mmarchini:
Daniel is on vacation, so I added a commit to address the nit.

mmarchini:
Landed in ecfbdca

PR #259, opened 2019-01-18
src: make source list -l flag optional

mmarchini:
`v8 source list` was expecting an undocumented flag -l, which is the
first line the command will output from source code. This commit makes
that flag optional, documents the flag, improves the help message for
this command and adds tests for it.

Fixes: https://github.com/nodejs/llnode/issues/138

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,7 +262,11 @@ The following subcommands are supported:
       print           -- Print short description of the JavaScript value.
 
                          Syntax: v8 print expr
-      source          -- Source code information
+      source list     -- Print source lines around the currently selected 
+                         JavaScript frame.
+                         Syntax: v8 source list [flags]
+                         Flags:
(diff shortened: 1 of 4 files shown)


mmarchini:
I tried #139 but for some reason, it didn't work, so this is an alternative to that PR.

cc @No9 @joyeecheung

README.md
```diff
@@ -262,7 +262,11 @@ The following subcommands are supported:
       print           -- Print short description of the JavaScript value.
 
                          Syntax: v8 print expr
-      source          -- Source code information
+      source list     -- Print source lines around the currently selected
```
joyeecheung: Do you remember why there are multiple words in this command? @indutny

mmarchini: I imagine it's because lldb has a `source list` command with a similar purpose.

mmarchini:
Landed in 0c1e3ae

PR #260, opened 2019-01-18
Run linter (clang-format) on Travis

mmarchini:


scripts/linter.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+
+const { getNativeBinary } = require("clang-format");
```
joyeecheung: If we are using the clang-format npm package, can we just use the `git-clang-format` script there with something like `git-clang-format --diff --quiet master src/*` and check the stdout?

mmarchini: There's no way to get `git-clang-format` path from their JS API though, but I guess we could use `npx git-clang-format` in this script. We could have `test -z "$(npx git-clang-format --diff --quiet master src/* | grep -v 'no modified files to format')"` in our package.json instead of this script, but this wouldn't work on Windows.

Overall I don't think llnode's code base is large enough to be worth using `git-clang-format` instead of plain `clang-format`, and most of the code base is already formatted correctly (and this PR fixes any leftover unformatted code), but I don't mind changing to `git-clang-format` either, as long as we have linter in Travis.

mmarchini: By the way, the output of `scripts/linter.js` right now is this:

```
The following files are not formatted correctly:
  llscan.cc
  llscan.h
```

It's a compact output, but on the other hand the goal is for the users to run `npm run format`, so it's not particularly useful and exiting with status code != 0 would be enough (maybe with a message ``Please run `npm run format`.``).

joyeecheung:
Ping @mmarchini This needs a rebase

mmarchini:
Done. I also moved `clang-format` dependency to `devDependencies`.

mmarchini:
Landed in ea06ef5...85f067a

PR #264, opened 2019-01-24
build: bump c++ to c++14

mmarchini:


mmarchini:
We'll need to drop support to Ubuntu 14.04 to enable `c++14`.

mmarchini:
Let's try again in the future.

PR #271, opened 2019-02-25
src: inspcet arguments from an Error object stack

mmarchini:
This is useful when the process aborts later than a thrown error (for
example, when there's an async rethrow or an unhandled Promise
rejection). V8 changes to make this possible landed in
https://github.com/v8/v8/commit/3724a125.

### DIFF
--- a/src/llv8.cc
+++ b/src/llv8.cc
@@ -1259,15 +1259,10 @@ StackTrace::StackTrace(JSArray frame_array, Error& err)
 
   len_ = v8::Smi(maybe_stack_len).GetValue();
 
-  multiplier_ = 5;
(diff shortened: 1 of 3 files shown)


mmarchini:
Still need to add some tests here. Also, there's no way to test this feature with current Node.js release lines, so we might want to wait to land it.

cjihrig:
Typo in the commit message: inspcet -> inspect

mmarchini:
Recent-ish changes on V8 rendered this feature less useful (it's less likely users will get an Error object with the raw stack we expect), so I'm closing it for now. We can revisit it if situation changes.

PR #272, opened 2019-02-25
src: add rudimentary Promise support

mmarchini:
This patch allows llnode to list Promise objects with findjsobjects,
findjsinstances and findrefs. We can investigate more fancy Promise
features in the future (such as listing handlers, Promise status, etc.).

### DIFF
--- a/src/llv8-constants.cc
+++ b/src/llv8-constants.cc
@@ -518,6 +518,13 @@ void Types::Load() {
   kLastContextType = LoadConstant("LastContextType");
 
   kJSErrorType = LoadConstant("type_JSError__JS_ERROR_TYPE");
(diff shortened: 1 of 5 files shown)


mmarchini:
The metadata landed in https://github.com/v8/v8/commit/8aaa186ed42, it will be available in Node.js v12. We can backport if we want this in Node.js v10.

cjihrig:
@mmarchini can you also update https://github.com/nodejs/node/blob/master/test/v8-updates/test-postmortem-metadata.js (or just ping me and I can do it).

mmarchini:
@cjihrig done https://github.com/nodejs/node/pull/31357

mmarchini:
Landed in 991c731deb42

PR #276, opened 2019-03-11
src: apply clang-format rules

mmarchini:
Since GitHub/Travis won't rerun tests for Pull Requests when `master` is updated, some commits landed after https://github.com/nodejs/llnode/commit/85f067aad8737d0d8891fd42b41694415f951a24, which made our linter test fail on `master. Also, since there are only a few open pull requests now, this is unlikely to happen again.


mmarchini:
cc @nodejs/llnode

mmarchini:
Landed in ecfbdca...62ca523

PR #285, opened 2019-07-08
src: reduce redundant "error:" for `inspect` and `findrefs`

oyyd:
When `inspect` or `findrefs` an invalid expr, llnode will print
an error message prefixed with "error: error: error:". This PR
will reduce one redundant "error:" for such scenario and it
seems that we can't remove the other redundant "error:" by
calling SB API only.

Before this PR:

```
(llnode) v8 inspect invalid_addr
error: error: error: use of undeclared identifier 'invalid_addr'
(llnode) v8 findrefs -v invalid_addr
error: error: error: use of undeclared identifier 'invalid_addr'
```

After:

```
(llnode) v8 inspect invalid_addr
error: error: use of undeclared identifier 'invalid_addr'
(llnode) v8 findrefs -v invalid_addr
error: error: use of undeclared identifier 'invalid_addr'
```

The other redundant "error:" seems to come from [here](https://github.com/llvm-mirror/lldb/blob/master/source/API/SBValue.cpp#L274) and I couldn't find a way to remove it by calling SB API only.

### DIFF
--- a/src/llnode.cc
+++ b/src/llnode.cc
@@ -192,10 +192,8 @@ bool PrintCmd::DoExecute(SBDebugger d, char** cmd,
   SBExpressionOptions options;
   SBValue value = target.EvaluateExpression(full_cmd.c_str(), options);
   if (value.GetError().Fail()) {
-    SBStream desc;
-    if (value.GetError().GetDescription(desc)) {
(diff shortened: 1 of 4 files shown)


mmarchini:
Landed in 74414fc

PR #291, opened 2019-09-17
src: fix JSError inspection with stringified stack

mmarchini:
When the `stack` property of an Error object is accessed in V8, a
stringified version of the stack is generated, and then the "raw stack"
(with FP and arguments) is removed from memory. Our current inspection
code couldn't handle this because it was not checking if the raw stack
was a valid object. This commit makes that code more robust and thus
fixes inspection of error objects with stringified stacks.

### DIFF
--- a/src/llv8.cc
+++ b/src/llv8.cc
@@ -1271,6 +1271,13 @@ StackTrace::Iterator StackTrace::end() {
 
 StackTrace::StackTrace(JSArray frame_array, Error& err)
     : frame_array_(frame_array) {
+  if (!frame_array.Check()) {
+    PRINT_DEBUG("JS Array is not a valid object");
+    len_ = -1;
+    multiplier_ = -1;
+    return;
+  }
+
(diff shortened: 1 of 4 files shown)


src/llv8.cc
```diff
@@ -1268,6 +1268,14 @@ StackTrace::Iterator StackTrace::end() {
 
 StackTrace::StackTrace(JSArray frame_array, Error& err)
     : frame_array_(frame_array) {
+  if (!frame_array.Check()) {
+    Error::PrintInDebugMode(
+        "JS Array is not a valid object");
```
cjihrig: Do you think this message would be more helpful if it mentioned that the invalid JS array is an error stack?

mmarchini: I thought about that, but I want to change our PrintInDebugMode to also print the function where it was called, so I decided to keep like this until I make that change.

mmarchini:
Landed in fb25c91

Issue #278, opened 2019-04-02
Flaky findrefs test on Node.js 10


mmarchini:
Has been happening a lot lately:

```
not ok 127 Should find reference #4
  ---
    operator: ok
    expected: true
    actual:   false
    at: sess.hasSymbol (/Users/travis/build/nodejs/llnode/test/plugin/scan-test.js:183:11)
    stack: |-
      Error: Should find reference #4
          at Test.assert [as _assert] (/Users/travis/build/nodejs/llnode/node_modules/tape/lib/test.js:226:54)
          at Test.bound [as _assert] (/Users/travis/build/nodejs/llnode/node_modules/tape/lib/test.js:77:32)
          at Test.assert (/Users/travis/build/nodejs/llnode/node_modules/tape/lib/test.js:344:10)
          at Test.bound [as ok] (/Users/travis/build/nodejs/llnode/node_modules/tape/lib/test.js:77:32)
          at sess.hasSymbol (/Users/travis/build/nodejs/llnode/test/plugin/scan-test.js:183:11)
          at linesUntil (/Users/travis/build/nodejs/llnode/test/common.js:318:7)
          at SessionOutput.onLine (/Users/travis/build/nodejs/llnode/test/common.js:113:5)
          at SessionOutput.emit (events.js:189:13)
          at Socket.SessionOutput.stream.on (/Users/travis/build/nodejs/llnode/test/common.js:63:14)
          at Socket.emit (events.js:189:13)
  ...
```

mmarchini:
Not a flaky test, this is actually a bug with (apparently) `findrefs`.

PR #295, opened 2019-09-23
src: fix findrefs for context locals

mmarchini:
Previous code was too strict when hitting a context which it was unable
to read Locals information: it would stop looping through the contexts
even if we still had valid contexts in the queue. Changing it to loop
through all context and all Locals in a context should give us a more
reliable results.

Also, if a given object was found in a context Locals but we're unable
to read the Local name, we'll show it as ??? to indicate it's being
referenced but we couldn't read its name (in case of corrupted memory,
for example).

Fixes: https://github.com/nodejs/llnode/issues/278


mmarchini:
Landed in 0dbfad0

PR #296, opened 2019-09-23
doc: update collaborator guide to match the project size

mmarchini:
This commits modifies our collaborator guide to match the current
project size. llnode is way smaller than nodejs/node, and requiring two
approvals to land a change/one approval with a seven day wait can be
quite bureaucratic. Two approvals represent 50% of our active reviewers
at the moment, which is quite a lot. Changing the requirements to one
approval with no wait time once a PR gets approved makes more sense for
this project, and it also matches other smaller projects across the
organization.

### DIFF
--- a/COLLABORATOR_GUIDE.md
+++ b/COLLABORATOR_GUIDE.md
@@ -1,4 +1,11 @@
 # llnode Collaborator Guide
 
 llnode follows the
-[Node.js Collaborator Guide](https://github.com/nodejs/node/blob/master/COLLABORATOR_GUIDE.md).
+[Node.js Collaborator Guide](https://github.com/nodejs/node/blob/master/COLLABORATOR_GUIDE.md),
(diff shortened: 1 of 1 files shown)


mmarchini:
cc @nodejs/llnode and @nodejs/diagnostics

mmarchini:
If this turns out to be a bad idea we can always go back to the previous policy later.

joyeecheung:
I wonder do we even need two approvals here? As far as I know most projects in this organization don't have this kind of requirement, and mostly operate on 1-apporval-whenever.

mmarchini:
I'm open to 1-apporval-whenever if others are.

mmarchini:
Updated COLLABORATORS_GUIDE with 1-apporval-whenever. Please let me know what y'all think :)

mmarchini:
Landed in afaec48

PR #300, opened 2019-09-27
src: use new SystemPointerSize constant

mmarchini:
This alone should fix a lot of issues. It was enough to get a reasonable
stack trace from a Node.js v12.3.0 core dump, ran with
--interpreted-frames-native-stack. For some reason `v8 bt` is not
working when running Node.js in lldb.


mmarchini:
Landed in 2c4c99c

PR #301, opened 2019-09-30
src: Treat embeded builtins as V8 functions

mmarchini:
On Node.js v12 most builtins are compiled during build time and embeded
in the binary. Because of that, LLDB will think it knows how to handle
these frames (especially JavaScript frames). To correctly handle those
frames, we check the function name, if it starts with `Builtins_` we'll
handle it as a JIT function. This method should be safe since any C++
function coming from V8 or Node.js will be mangled, thus beginning with
`_Z...`. There might be a better way to handle this, but for now this
check should be enough.

### DIFF
--- a/src/llnode.cc
+++ b/src/llnode.cc
@@ -69,7 +69,12 @@ bool BacktraceCmd::DoExecute(SBDebugger d, char** cmd,
     const char star = (frame == selected_frame ? '*' : ' ');
     const uint64_t pc = frame.GetPC();
 
-    if (!frame.GetSymbol().IsValid()) {
+    // TODO(mmarchini): There might be a better way to check for V8 builtins
+    // embeded in the binary.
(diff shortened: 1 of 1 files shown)


mmarchini:
cc @nodejs/llnode

src/llnode.cc
```diff
@@ -69,7 +69,12 @@ bool BacktraceCmd::DoExecute(SBDebugger d, char** cmd,
     const char star = (frame == selected_frame ? '*' : ' ');
     const uint64_t pc = frame.GetPC();
 
-    if (!frame.GetSymbol().IsValid()) {
+    // TODO(mmarchini): There might be a better way to check for V8 builtins
+    // embeded in the binary.
```
cjihrig:
```suggestion
    // embedded in the binary.
```

mmarchini:
Landed in 2964af5

PR #302, opened 2019-09-30
src: fix String::Length for Node.js v12.3.0

mmarchini:
Type of String::Length changed from SMI to int32. This commit changes
String::Length to use the new type. String::Length now returns a
CheckedType, and places calling String::Length will check the type using
.Check instead of Error::Fail, making the code more resilient in case
the String length can't be loaded properly.

### DIFF
--- a/src/llv8-constants.cc
+++ b/src/llv8-constants.cc
@@ -299,7 +299,12 @@ void String::Load() {
   kExternalStringTag = LoadConstant("ExternalStringTag");
   kThinStringTag = LoadConstant("ThinStringTag");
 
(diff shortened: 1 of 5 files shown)


mmarchini:
For posterity, the SMI to int32 change happened here: https://chromium-review.googlesource.com/c/v8/v8/+/1224432/

I'm not sure this pull request fixes all the problems though. I'm still seeing a lot of frames not loading with `v8 bt`...

mmarchini:
cc @nodejs/llnode

mmarchini:
Landed in a79a008

Issue #255, opened 2018-12-21
Upcoming metadata changes in V8 7.2


cjihrig:
The V8 7.2 update requires the following adjustments to the postmortem debugging metadata constants:

- `v8dbg_prop_idx_first` was removed in v8/v8@1ad0cd5

- `v8dbg_jsarray_buffer_was_neutered_mask` was renamed to `v8dbg_jsarray_buffer_was_detached_mask` in v8/v8@f68ee6e

- `v8dbg_jsarray_buffer_was_neutered_shift` was renamed to `v8dbg_jsarray_buffer_was_detached_shift` in v8/v8@f68ee6e

- `v8dbg_class_Map__instance_descriptors__DescriptorArray` moved to `v8dbg_class_Map__raw_instance_descriptors__DescriptorArray` and began using `ACCESSORS2` in v8/v8@799dfad

The following postmortem debugging constants were also impacted
by V8's introduction of `ACCESSORS2`:

- `v8dbg_class_ThinString__actual__String` changed in v8/v8@0f581e4

- `v8dbg_class_UncompiledData__inferred_name__String` changed in v8/v8@0f581e4

- `v8dbg_class_JSFunction__shared__SharedFunctionInfo` changed in v8/v8@8162090

- `v8dbg_class_SharedFunctionInfo__function_data__Object` had its accessor removed in v8/v8@a55803a. This will be fixed in gen-postmortem-metadata.py.

Refs: https://github.com/nodejs/node-v8/pull/96

mmarchini:
FYI we'll have to add some constants for DescriptorArray now that it is a HeapObject and not a FixedArray. I'll send a PR to V8 today or tomorrow.

cjihrig:
@mmarchini if you add new constants, can you also PR them to Node's postmortem metadata test? I think that makes it a lot easier to know when things are going to break. Alternatively, if you don't want to update the test, just point me to the new constants and I can add them myself.

mmarchini:
I'll update the metadata test once https://chromium-review.googlesource.com/c/v8/v8/+/1832311 lands on core :+1:

PR #303, opened 2019-10-01
src: introduce Constant class

mmarchini:
Same idea as CheckedType: instead of relying on the default value for
constants to determine if they were loaded correctly or not (default
value usually was -1), have a class which will have information about
the loaded constant. This can help us:

  - Check if a constant was properly loaded before using it (otherwise
    we'd get into undefined behavior, and there are a lot of places
    where this happens.
  - Check if a constant value was loaded or if we're looking at the
    default value.
  - Explicitly define a constant as optional (constants which are not
    relevant except for specific V8 versions).

This will help us improve reliability and debugability of llnode.


mmarchini:
I'll update the commit message later, but the main reason I started to work on this was to reduce noise from LookupConstant and LoadConstant debug messages. We don't want to see a message if a constant wasn't loaded but the replacement constant was (we were seeing two messages for the same constant in this case). The reduced noise from this PR helped me a lot while working on #255 (PR for that will come after this one lands).

mmarchini:
cc @nodejs/llnode

mmarchini:
Ping @nodejs/llnode. This is the main blocker for the patches I have to fix inspection of Node.js v12 core dumps/processes.

(I can rewrite those patches in the old way if this `Constant` class is not something we want, but from experience the code looks cleaner and easier to debug with this patch)

mmarchini:
Ping @nodejs/llnode @nodejs/diagnostics @nodejs/post-mortem

mmarchini:
* Changed LookupConstant to return a Constant object instead of `std::pair<int64_t, bool>`
* Added a ConstantStatus to replace `bool loaded_` and `bool valid_`
* Added a comment explaining how the Constant class should be interpreted and used
* Moved Constant constructors to protected and friended `Constants` to prevent misuse of the class (it should only be used to load constants).

As before, I kept the original LoadRawConstant and LoadConstant signatures unchanged to avoid a major refactor of `llv8-constants.cc` in a single PR.

mmarchini:
Landed in 2afd59e

PR #308, opened 2019-10-03
tools: git ignore core dumps

mmarchini:


mmarchini:
Landed in 7b9598e9dad0

PR #311, opened 2019-10-03
src: fix function name lookup for inferred names

mmarchini:
Apparently, sometimes the FunctionName slot on ScopeInfo is filled with
the empty string instead of not existing. This commit changes our
heuristic to search for the first non-empty string on the first 3 slots
after the last context info slot on the ScopeInfo. This should be enough
to cover most (all?) cases.

Also updated frame-test to add frames to the stack which V8 will infer
the name instead of storing it directly, and changed this particular
test to use Promises instead of callbacks. We should be able to upgrade
tests to primise-based API gradually with this approach. When all tests
are promisified, we can change the api on test/common.js to be
promise-based instead of callback-based.

### DIFF
--- a/test/fixtures/frame-scenario.js
+++ b/test/fixtures/frame-scenario.js
@@ -3,13 +3,29 @@ const common = require('../common');
 
 function crasher(unused) {
   'use strict';
-  process.abort();  // Creates an exit frame.
-  return this;      // Force definition of |this|.
+  this.foo = arguments;  // Force adaptor frame on Node.js v12+
+  process.abort();       // Creates an exit frame.
+  return this;           // Force definition of |this|.
 }
 
-function eyecatcher() {
+// Force V8 to use an inferred name instead of saving the variable name as
+// FunctionName.
+let fnInferredName;
+fnInferredName = (() => function () {
   crasher();    // # args < # formal parameters inserts an adaptor frame.
+})();
+
+function Module() {
+  this.foo = "bar";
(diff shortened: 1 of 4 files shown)


mmarchini:
I wonder if the failing tests are happening because I changed the test to use Primises... llnode is printing `v8 source list` and `v8 source list -l 2` correctly, but the test suite is not receiving it sometimes...

test/fixtures/frame-scenario.js
```diff
@@ -3,13 +3,29 @@ const common = require('../common');
...
+  return this;           // Force definition of |this|.
 }
 
-function eyecatcher() {
+// Force V8 to use an inferred name instead of saving the variable name as
+// FunctionName.
+let fnInferredName;
+fnInferredName = (() => function () {
   crasher();    // # args < # formal parameters inserts an adaptor frame.
+})();
```
mmarchini: V8 is not inferring this on Node.js v12... We'll probably need to skip the check for this function on 12.

Upstream bug: https://bugs.chromium.org/p/v8/issues/detail?id=9807

mmarchini:
cc @nodejs/llnode

mmarchini:
cc @nodejs/llnode

mmarchini:
Ping @nodejs/llnode, we need this for v12 support.

mmarchini:
Landed in fd9d2b099093...8068cdad28f4

PR #313, opened 2019-10-03
src: use Check() instead of raw != -1

mmarchini:
a501635516aa0c153b1e17b62e47150b6458af99 changed the behavior of Check()
so that it could be properly used to determine if an object was loaded
successfully from memory. In the past we used raw != -1 to perform the
same check. Some places were still using the old approach, which has
less guarantees than Check(). This commit changes those places to use
the new approach. Also added a few RETURN_IF_THIS_INVALID guards on the
functions touched by this commit.


mmarchini:
cc @nodejs/llnode

mmarchini:
Landed in 37b3c2684432

PR #314, opened 2019-10-05
reword! allow use of accessors on v8 inspect and print

mmarchini:
Still a work in progress, but this should make it easier to inspect objects:

image

Still have a few things to do:

- [ ] Validate boundary of accessors
- [ ] Improve error messages
- [ ] Move the address+accessor parsing outside of the Cmd function
- [ ] Use strings as accessors (like `['large-array']`) (could be a follow-up PR)

### DIFF
--- a/src/llnode.cc
+++ b/src/llnode.cc
@@ -4,6 +4,8 @@
 #include <string.h>
 
 #include <cinttypes>
+#include <iostream>
+#include <regex>
 #include <sstream>
 #include <string>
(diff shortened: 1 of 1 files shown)


mmarchini:
I'm also looking into completion for accessors, but this requires changes to LLDB, so it will take some time until we can use it on llnode.

mmarchini:
Closing, might get back to it eventually

PR #315, opened 2019-10-07
build: allow custom lldb headers path with npm install

mmarchini:
It's possible to set a custom lldb headers path if configuring the build
manually. This commit also allows users to do it when using
`npm install`. This is useful when using lldb built from source, for
example.


mmarchini:
Landed in 9a58d8cf3fa1

PR #323, opened 2019-11-07
build: move CI to GitHub Actions

mmarchini:
Reduce CI time by moving to GitHub Actions.


.github/workflows/push.yml
```diff
@@ -0,0 +1,45 @@
+name: on push or pull_request
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.ox }}
```
richardlau:
```suggestion
    runs-on: ${{ matrix.os }}
```
?

mmarchini: Oops

richardlau:
Over in CITGM we're proposing keeping Travis around for code coverage: https://github.com/nodejs/citgm/pull/759

mmarchini:
> Over in CITGM we're proposing keeping Travis around for code coverage: nodejs/citgm#759

Let me take a look, I don't think we had to enable any permissions for codecov in this repository (I remember making a PR and codecov was just working).

richardlau:
> 
> 
> > Over in CITGM we're proposing keeping Travis around for code coverage: [nodejs/citgm#759](https://github.com/nodejs/citgm/pull/759)
> 
> Let me take a look, I don't think we had to enable any permissions for codecov in this repository (I remember making a PR and codecov was just working).

I think that was the point @targos was making -- that on Travis nothing extra had to be enabled for codecov but that wasn't the case for Actions.

richardlau:
For reference: https://community.codecov.io/t/whitelist-github-action-servers-to-upload-without-a-token/491

mmarchini:
fyi I'm testing on my fork to avoid notification spam here.

I'm having trouble getting tests to pass because apparently core dumps are not being generated. Not sure why yet, based on system settings it doesn't look like core dumps are blocked.

mmarchini:
Should be working now. A few things we still need to do:

- [ ] Daily runs
- [ ] Update README
- [ ] Code coverage. codecov said they will enable GITHUB_TOKEN support after Nov 13 (https://community.codecov.io/t/whitelist-github-action-servers-to-upload-without-a-token/491/8). Still won't work for forks though. TBH I don't feel strong enough about keeping codecov around if we need two CI services for that (llnode needs to run all tests on all supported Node.js versions to generate a relevant coverage report, keeping travis would kill the purpose of speeding up CI).
- [ ] Better handling of nightly/canary tests (I have ideas, will probably open an issue about that later)

mmarchini:
Ok, I'm happy with this setup now. CI is a lot faster than Travis, and it's able to run more jobs in parallel as well.

We lost the build matrix for now, but I don't think it's a problem. If someone creates a https://github.com/bjfish/travis-matrix-badges for GitHub Actions, we can use it. I also wouldn't be surprised if GitHub makes this information easily accessible in the repository home page in the future.

I had to change coverage from Codecov to Coveralls because Codecov doesn't use the `GITHUB_TOKEN` yet. Coveralls doesn't add a comment in the repository, but coverage information is available as a GitHub Check:

img-2020-01-09-103941

cc @nodejs/llnode @richardlau

.github/workflows/push.yml
```diff
@@ -0,0 +1,74 @@
...
+
+jobs:
+  tests:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        node:
+          - version: 8.x
```
richardlau: Node.js 8 is now End-of-Life so I'd have no objections to dropping it from the matrix. Same with Node.js 11 (which should probably be replaced with 12 and 13).

Otherwise LGTM.

mmarchini: Yes, I'll do it in a follow up.

mmarchini:
Landed in 9a58d8cf3fa1...ecb8d0e34236

PR #324, opened 2019-11-08
src: update UncompiledData types constants

mmarchini:
Ref: https://github.com/nodejs/llnode/issues/255


mmarchini:
cc @nodejs/llnode

mmarchini:
Landed in fde0c59cdabd

PR #326, opened 2019-12-10
test: make func source test compatbile with v12

mmarchini:
The string stored in V8 with the function source code is different on
v12. Update the regexp used in our test to match older and newer
versions.


mmarchini:
cc @nodejs/llnode

mmarchini:
Ping @nodejs/llnode, we need this for v12 support.

mmarchini:
Landed in fd9d2b099093

PR #327, opened 2019-12-10
test: update Error.stack test for V8 7.5

mmarchini:
V8 7.5 changed how it caches the stringified stack on Error objects
after the first access. Instead of replacing the accessor Error.stack
with the stringified stack, the stringified stack is stored in the same
Symbol which was used to store the FrameArray stack.

Ref: https://github.com/v8/v8/commit/c8206043e1afb6d179a68bb5a8c079de5e


mmarchini:
Landed in f7f9347b19e1

PR #329, opened 2020-01-14
build: add project ncurc config

mmarchini:


mmarchini:
Landed in 878b514e2919

PR #330, opened 2020-01-16
Nodejs v12 support

mmarchini:
This PR contains all commits necessary to make llnode work on v12, plus changes to CI to run tests on v12, drop v8 and v11, and bump the major version since we're dropping support and adding new one. I made a PR on nodejs/node to update the list of postmortem constants used, but I noticed some of the constants here already changes on v13, so we'll have to work on that.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -16,9 +16,8 @@ jobs:
       fail-fast: false
       matrix:
         node:
-          - version: 8.x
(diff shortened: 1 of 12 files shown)


mmarchini:
cc cc @nodejs/llnode @nodejs/diagnostics

mmarchini:
Landed in 878b514e2919...e8896e0d3932

PR #332, opened 2020-01-16
src: harden SlicedString::ToString

mmarchini:
Add some extra checks to make sure we won't crash when stringifying a
(possibly corrupted) SlicedString.


mmarchini:
I had this issue in the wild. Couldn't come up with a test case for it though...

mmarchini:
Landed in bff9f5ea3e75

Issue #242, opened 2018-10-05
Landing PRs policy


mmarchini:
I just realized we don't have a policy for landing Pull Requests and I guess most of us are assuming Node.js core policies here: at least one approval + wait 48/72 hours. Considering the project has only a few collaborators, I don't think this policy is adequate since it usually takes weeks to get approvals.

Inspired on recent discussions about PR waiting time on Node.js core (https://github.com/nodejs/node/pull/22275) and on the [Commit-Then-Review](https://www.apache.org/foundation/glossary.html#CommitThenReview) concept, I'd like to propose the following PR policy:

```
Before landing a pull request, make sure the following conditions must be true:
- The Pull Request must contain tests, when applicable
- Last CI run must be green
- There must be no objections (GitHub's Request for Changes)

Also, at least one of the following conditions must be true:
- The Pull Request is open for at least 72 hours
- The Pull Request was approved by at least one collaborator
```

I'm open to discuss this in person next week during the Collab Summit :)

cc @nodejs/diagnostics since llnode falls under the WG Governance. Also ccing llnode collaborators: @joyeecheung @indutny @cjihrig @bnoordhuis @rnchamberlain @hhellyer

richardlau:
The one advantage of following the same rules as core is that the associated tooling (i.e. `node-core-utils`) should just work. That's not to say we couldn't teach ncu different rules for different projects (but then we'd have to make sure it applied the correct rules).

mmarchini:
`node-core-utils` already have problems with llnode: it complains because there's no CI run (we're not using Jenkins yet) and users have to use `ncu-config` to make ncu look into Node.js README for the list of collaborators (which doesn't make sense, llnode collaborators are not necessarily Node.js collaborators). Thus I don't think this should be a reason to keep following core's policy.

By the way, I made a mistake in my first comment: right now core requires two approvals instead of one, unless the pull request is open for over a week. This rule works for core where we have several collaborators who review pull requests daily, but it is excessive for a project with only a handful of active collaborators.

----

With that being said, teaching ncu different rules according to the project sounds like a good idea. Maybe we could have a .ncurc in our projects, where we can customize ncu's behavior accordingly?

For example:

```
{
  "wait-time": 72,
  "min-approvals": 2,
  "required-ci": [ 
    { "type": "jenkins", "job": "node-test-pull-request" }, 
    { "type": "jenkins", "job": "node-test-pull-request-lite-pipeline" }
  ]
}
```

[...]

PR #336, opened 2020-02-12
meta: add policy to land PRs without approval

mmarchini:
As discussed in the 2020-02-12 Diagnostics WG Meeting and suggested in
https://github.com/nodejs/llnode/issues/242#issuecomment-430369666,
change the landing policy so that Collaborators can land pull requests
if it was open for three days and there were no reviews on it.

Ref: https://github.com/nodejs/diagnostics/issues/355

### DIFF
--- a/COLLABORATOR_GUIDE.md
+++ b/COLLABORATOR_GUIDE.md
@@ -9,3 +9,6 @@ with the following differences:
   - There's no minimum wait time for landing a pull request. Pull requests can
     land as soon as they get approved by at least one Collaborator.
   - Approval must be from Collaborators who are not authors of the change.
+  - If the pull request was open for three days (72 hours) without reviews, and
+    it was opened by a llnode Collaborator, the pull request can land without
+    approvals.


mmarchini:
cc @nodejs/diagnostics @nodejs/llnode @nodejs/post-mortem please let me know if there are any objections

COLLABORATOR_GUIDE.md
```diff
@@ -9,3 +9,6 @@ with the following differences:
   - There's no minimum wait time for landing a pull request. Pull requests can
     land as soon as they get approved by at least one Collaborator.
   - Approval must be from Collaborators who are not authors of the change.
+  - If the pull request was open for three days (72 hours) without reviews, and
```
cjihrig: Not to intentionally slow llnode down any further, but what about 7 days to match the one approval clause from https://github.com/nodejs/node/blob/master/COLLABORATOR_GUIDE.md#code-reviews.

mmarchini: I'm not entirely opposed to 7 days, but I don't think "to match the one approval clause from core" is a good reason to do so. I'm open to hearing other arguments to use 7 days instead of 3 though.

On a counter-argument to 7 days, if there is a bug and a PR with a fix doesn't get a review, it will now take seven days to release a fix. We could have a shorter waiting period for bug fixes, but that seems to overly complicate the policy.

cjihrig: To counter your counter-argument, if there is a bug under this proposal, it will take 3 days to release a fix 😄. I guess any amount of waiting will be an arbitrary amount. I only recommended going with 7 days because that was the agreed upon answer to this exact same problem in core. At the end of the day, you're the one doing literally all of the work, and I'm not going to block this PR.

mmarchini:
Keeping this open until tomorrow afternoon (PST) in case there are any objections from folks who didn't comment yet

mmarchini:
Landed in bd3a1c7eac4c

PR #339, opened 2020-02-14
tools: add script with commands to facilitate dev

mmarchini:
Some tasks are very common when working on llnode, especially when
upgrading a V8 version. This new script is intended to make developers
life easier by simplifying tasks which usually require printfs on llnode
code or manually inspecting the memory of a process. Hopefully we'll add
more commands in the future.


mmarchini:
Landed in 1948512b5e71

PR #344, opened 2020-03-05
doc: use official CI badge

gengjiawen:


gengjiawen:
> Humm, will check tomorrow, but the official one seems to be for the last status regardless if the build was on a PR or on master, whereas the shields badge will get the last result only from master? Otherwise lgtm

The shields version is broken, not sure why. As the icon page suggest the badge is for the default branch. You can lock to master branch too.

mmarchini:
Oh, nice catch. There was a typo on the Shields URL (`push` instead of `on%20push%20or%20pull_request`). Do you mind updating the URL to keep using shield (so that all three badges look the same)? The correct url should be `https://img.shields.io/github/workflow/status/nodejs/llnode/on%20push%20or%20pull_request/master?style=flat-square`

No9:
@gengjiawen 
Apologies for having this open for so long but now we have landed the work to support 16/18 I am looking to clean up the outstanding PRs.
Do you want to sync your PR and apply the final change that Mary requested? 
i.e. apply the flat-style property 
https://img.shields.io/github/workflow/status/nodejs/llnode/on%20push%20or%20pull_request/master?style=flat-square 

Happy to just make this change myself if it's too much bother but wanted to capture your work if you are still interested.

Issue #346, opened 2020-03-25
 crash when run v8 i -l


Jamy325:
```
(llnode) v8 i 0x11af8ff2f819                                        
0x11af8ff2f819:<Array: length=100 {
    [0]=0x124351ab1901:<Object: Object>,
    [1]=0x124351ab1949:<Object: Object>,
    [2]=0x124351ab1991:<Object: Object>,
    [3]=0x124351aafec9:<Object: Object>,
    [4]=0x124351aaff11:<Object: Object>,
    [5]=0x124351aaff59:<Object: Object>,
    [6]=0x124351aaffa1:<Object: Object>,
    [7]=0x124351aaffe9:<Object: Object>,
    [8]=0x124351ab0031:<Object: Object>,
    [9]=0x124351aaed01:<Object: Object>,
    [10]=0x124351aaed49:<Object: Object>,
    [11]=0x124351aaed91:<Object: Object>,
    [12]=0x124351aaedd9:<Object: Object>,
    [13]=0x124351aaee21:<Object: Object>,
    [14]=0x124351aaeb41:<Object: Object>,
    [15]=0x124351aaeb89:<Object: Object>}>

(llnode) v8 i -l 20 0x11af8ff2f819
Stack dump:
0.	Program arguments: /usr/local/bin/lldb --one-line plugin load '/opt/node_modules/llnode/llnode.so' --one-line settings set prompt '(llnode) ' ./node -c n.core.46706 
 #0 0x000000000041daca llvm::sys::PrintStackTrace(llvm::raw_ostream&) (/usr/local/bin/lldb+0x41daca)
 #1 0x000000000041be7a llvm::sys::RunSignalHandlers() (/usr/local/bin/lldb+0x41be7a)
 #2 0x000000000041c25c SignalHandler(int) (/usr/local/bin/lldb+0x41c25c)
 #3 0x00007fc1ef3067e0 __restore_rt (/lib64/libpthread.so.0+0xf7e0)
 #4 0x00007fc1e78dbe72 ____strtoll_l_internal (/lib64/libc.so.6+0x36e72)
 #5 0x00007fc1e41545f6 llnode::ParsePrinterOptions(char**, llnode::Printer::PrinterOptions*) /opt/zm5/game-admin-web/node_modules/llnode/build/../src/llscan.cc:76:0
[...]

PR #347, opened 2020-03-25
src: fix getopt format string

bnoordhuis:
`v8 inspect -l <n>` takes an argument.

Bug introduced in commit 2fc1571 ("src: option to limit output
of `findjsinstances`") from September 2018.

Fixes: https://github.com/nodejs/llnode/issues/346

### DIFF
--- a/src/llscan.cc
+++ b/src/llscan.cc
@@ -62,7 +62,7 @@ char** ParsePrinterOptions(char** cmd, Printer::PrinterOptions* options) {
   optind = 0;
   opterr = 1;
   do {
-    int arg = getopt_long(argc, args, "Fmsdvln:", opts, nullptr);
+    int arg = getopt_long(argc, args, "Fmsdvl:n:", opts, nullptr);
     if (arg == -1) break;
 
     switch (arg) {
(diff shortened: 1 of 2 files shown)


mmarchini:
We should add some testing here. Since we don't seem to have specific tests for command arguments, we can change one of the inspect tests:

```diff
diff --git test/plugin/inspect-test.js test/plugin/inspect-test.js
index 0a59c9b..f664237 100644
--- test/plugin/inspect-test.js
+++ test/plugin/inspect-test.js
@@ -244,7 +244,7 @@ const hashMapTests = {
       });
     }, (t, sess, addresses, name, cb) => {
       const address = addresses[name];
-      sess.send(`v8 inspect --array-length 1 ${address}`);
+      sess.send(`v8 inspect -l 1 ${address}`);
 
       sess.linesUntil(/\]>/, (err, lines) => {
         if (err) return cb(err);
```

We use `--array-length` on other places in this test, so it should be fine.

Also our documentation is incomplete, `--array-length` and `--string-length` are not documented, but this can be handled in a separate PR.

bnoordhuis:
I updated one of the tests. I should perhaps mention that I don't plan to spend much more time on this patch, I just sent it because I'm such a nice guy (and also because obvious bug is obvious.)

mmarchini:
Landed in 83c810faa06d

PR #351, opened 2020-03-28
src: improve Error message on LoadPtr

mmarchini:
Knowing why LoadPtr failed can be quite helpful when debugging llnode,
but the current message doesn't give any additional information. Sharing
the address llnode tried to load, as well as the supposed type for that
address makes debugging easier. The type name is only shown on Debug
builds.


mmarchini:
Landed in 461e83aa0c54

PR #352, opened 2020-03-28
src: update SFI script accessor for V8 8.1

mmarchini:
The metadata used to access the script field from an SFI changed on V8
8.1. The field is still the same, so no other changes are necessary.
This fixes `v8 bt` on 8.1, as well as `stack-test` suite.


mmarchini:
Landed in 8e2a55c82e99

PR #354, opened 2020-03-28
src: fix some warnings and erroneous PRINT_DEBUG

mmarchini:


mmarchini:
Landed in 364e0346863c

PR #356, opened 2020-04-02
doc: fix supported version comment on README

mmarchini:


mmarchini:
I don't think it should be a semver major since we dropped support before, but I understand if other folks feel otherwise.

Also, the text is not entirely accurate, since v13 is not supported (and probably never will be).

mmarchini:
Landed in fdddce0d2c7d

Issue #355, opened 2020-04-02
npm install fails with Ubuntu 18.04, lldb 6.0


jlevon:
```
e$ node-gyp -v
v3.6.2
~$ dpkg -l lldb lldb-6.0-dev
ii  lldb                                    1:6.0-41~exp5~ubuntu1    amd64                    Next generation, high-performance debugger
un  lldb-6.0-dev                            
```

fails "npm install" (w or without --debug):

```
$ npm install --debug

> llnode@3.1.0 preinstall /home/jlevon/src/llnode
> node scripts/configure.js

Build dir is: /home/jlevon/src/llnode
Looking for lldb executable...
Found lldb executable /usr/bin/lldb

Reading lldb version...
Installing llnode for /usr/bin/lldb, lldb version 6.0

Looking for llvm-config for lldb 6.0...
Using llvm-config in /usr/bin/llvm-config-6.0

Looking for headers for lldb 6.0...
Found lldb headers in /usr/lib/llvm-6.0/include

Looking for shared libraries for lldb 6.0...
Found liblldb.so in /usr/lib/llvm-6.0/lib
Writing config.gypi:
{
  "variables": {
    "lldb_lib_dir%": "/usr/lib/llvm-6.0/lib",
    "lldb_lib%": "lldb",
    "lldb_include_dir%": "/usr/lib/llvm-6.0/include",
    "build_addon": "false",
    "coverage": "false"
  }
}
Writing llnode.js shortcut to /home/jlevon/src/llnode/llnode.js

> llnode@3.1.0 install /home/jlevon/src/llnode
> node scripts/install.js

make: Entering directory '/home/jlevon/src/llnode/build'
  CXX(target) Debug/obj.target/plugin/src/constants.o
  CXX(target) Debug/obj.target/plugin/src/error.o
  CXX(target) Debug/obj.target/plugin/src/llnode.o
  CXX(target) Debug/obj.target/plugin/src/llv8.o
../src/llv8.cc:732:36: warning: trigraph ??) ignored, use -trigraphs to enable [-Wtrigraphs]
   RETURN_IF_INVALID(name, "Symbol(???)");
                                     
../src/llv8.cc: In member function ‘llnode::v8::Value llnode::v8::JSObject::GetDescriptorProperty(std::__cxx11::string, llnode::v8::Map, llnode::Error&)’:
../src/llv8.cc:1179:14: warning: variable ‘value’ set but not used [-Wunused-but-set-variable]
       double value;
[...]

PR #357, opened 2020-04-03
build: add node-gyp as a dependency

mmarchini:
Fixes: https://github.com/nodejs/llnode/issues/355

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
   },
   "dependencies": {
     "bindings": "^1.3.0",
-    "node-addon-api": "^1.1.0"
+    "node-addon-api": "^1.1.0",
+    "node-gyp": "^6.1.0"
   }
 }


bnoordhuis:
As a maintainer of node-gyp, depending on a specific version of node-gyp is something I advise against. You should normally use the one that's bundled with node+npm.

mmarchini:
Thanks for the feedback @bnoordhuis, I'll probably not merge this then.

Any suggestions on #355? Or is it a situation we just can't handle in code, since it's caused by Ubunut/Debian distributing a non-official Node.js package?

bnoordhuis:
#355 seems like a problem of debian or ubuntu's own making. If those distros tracked v8.x LTS like they should have, they'd be at a node-gyp version that's new enough (v5.0.5 vs v3.6.2.)

mmarchini:
Closing in favor of https://github.com/nodejs/llnode/pull/367

Issue #353, opened 2020-03-28
Double-check unused Double Field code


mmarchini:
The if branch on https://github.com/nodejs/llnode/blob/master/src/llv8.cc#L1178-L1202 is essentially a noop: `double value` is set but never used, since we'll always return `Value()`. Not sure if we should be returning `double value` or if we can remove this code.

PR #358, opened 2020-04-06
src: boxed double fields for V8 8.1

mmarchini:
V8 8.1 disabled unboxed double fieds, which means all double fields are
now stored on memory as references to HeapNumber objects instead of the
double value. In theory V8 could store boxed doubles before, but we
didn't take it into account in our dobule fields code. In the future, V8
intends to re-enable unboxed doubles, which means we'll need to add
logic to handle both types of fields in the same llnode session.

This also fixes a bug where llnode was not handling double fields at all
in some cases.

Fixes: https://github.com/nodejs/llnode/issues/353

### DIFF
--- a/src/llv8-inl.h
+++ b/src/llv8-inl.h
@@ -89,6 +89,10 @@ inline bool HeapObject::Check() const {
          (raw() & v8()->heap_obj()->kTagMask) == v8()->heap_obj()->kTag;
 }
 
+inline bool HeapNumber::Check() const {
+  if (unboxed_double_) return Value::Check();
+  return HeapObject::Check();
+}
 
 int64_t HeapObject::LeaField(int64_t off) const {
   return raw() - v8()->heap_obj()->kTag + off;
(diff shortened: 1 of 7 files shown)


src/llv8-inl.h
```diff
@@ -388,8 +392,37 @@ inline T JSObject::GetInObjectValue(int64_t size, int index, Error& err) {
...
+  HeapObject map_obj = GetMap(err);
+  if (err.Fail()) HeapNumber();
+
+  Map map(map_obj);
+  int64_t instance_size = map.InstanceSize(err);
+  if (err.Fail()) return HeapNumber();
+
+  // TODO(mmarchini): Explain why index might be lower than zero.
+  if (index < 0) {
+    // When unbox doubles are not allowed, all double fields are stored as
```
addaleax, line 405:
```suggestion
    // When unboxed doubles are not allowed, all double fields are stored as
```

mmarchini:
I'm getting the OS X failures on master as well, and we're also seeing it on the daily runs: https://github.com/nodejs/llnode/runs/603630513?check_suite_focus=true. I think something changed on OS X lldb output when running a program, because that's where the test is failing. Will open an issue to investigate. In the meantime, llnode _should_ still work on OS X, but we don't have coverage until the issue is resolved.

PR #359, opened 2020-04-09
scripts: remove unwanted require

rickyes:


mmarchini:
Before every action started failing for no apparent reason, we had two failures on OS X which I think are not related to this PR (might be related to yesterday's Node.js releases): https://github.com/nodejs/llnode/actions/runs/74853245

PR #361, opened 2020-04-22
 src: update RegExp type constant for V8 8.1

mmarchini:
This is the last change needed to support Node.js v14.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -19,8 +19,9 @@ jobs:
           - version: 10.x
           - version: 12.x
           - version: 14.x
+          - version: 15.x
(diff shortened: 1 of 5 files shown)


.github/workflows/push.yml
```diff
@@ -19,8 +19,9 @@ jobs:
           - version: 10.x
           - version: 12.x
           - version: 14.x
+          - version: 15.x
```
mmarchini, line 22: Note this will fail for a few days until we bump the major version on core, but that's fine since those tests are allowed to fail anyway.

mmarchini:
Landed in dd57bfb8af3c...4918962beed7

Issue #362, opened 2020-04-22
lldb fails to attach to live process on OS X on CI


mmarchini:
https://github.com/nodejs/llnode/runs/603630513?check_suite_focus=true

Running on a local OS X machine which exhibits the same issues shows that lldb is failing to start a process and attach to it via `run`. This is a lldb issue but it is preventing us from running our tests on CI.

PR #368, opened 2020-08-05
chore(ci): temporarily disable OS X

mmarchini:
Refs: https://github.com/nodejs/llnode/issues/362

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -23,6 +23,9 @@ jobs:
             mirror: https://nodejs.org/download/nightly
           - version: 15.x
             mirror: https://nodejs.org/download/v8-canary
+        # os: [ubuntu-latest, macos-latest]
+        # Temporarily disable MacOS until
+        # https://github.com/nodejs/node/issues/32981 is fixed
         os: [ubuntu-latest]
     steps:
       - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


mmarchini:
cc @nodejs/llnode

joyeecheung:
It seems that `--abort-on-uncaught-exception` has regressed on Node.js v14 which is why the test is timing out (instead of aborting, the process exits with -1 normally)

.github/workflows/push.yml
```diff
@@ -23,7 +23,7 @@ jobs:
             mirror: https://nodejs.org/download/nightly
           - version: 15.x
             mirror: https://nodejs.org/download/v8-canary
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest]
```
joyeecheung, line 26:
```suggestion
        # os: [ubuntu-latest, macos-latest]
        # Temporarily disable MacOS until
        # https://github.com/nodejs/node/issues/32981 is fixed
        os: [ubuntu-latest]
```

mmarchini:
> It seems that `--abort-on-uncaught-exception` has regressed on Node.js v14 which is why the test is timing out (instead of aborting, the process exits with -1 normally)

Ah yes, that might cause v14 tests to fail, although I'm pretty sure llnode is broken on v14 since we bumped V8 recently (as in, after the last time we updated things on llnode).

joyeecheung:
From https://github.com/nodejs/node/issues/32981 it seems the bug was introduced from v12.11.0. Is there any progress in the V8 upstream?

mmarchini:
I don't think I opened an upstream bug :thinking: I'll check tomorrow and will open one if I haven't yet.

mmarchini:
Although tests are passing on v12, so that issue should be unrelated (hopefully)

PR #371, opened 2020-08-05
chore(ci): use codecov instead of coveralls

mmarchini:
coveralls is failing, and codecov seems to be working on actions now, so
replace coveralls with codecov.


mmarchini:
@nodejs/llnode @nodejs/diagnostics this unbreaks CI on pull requests.

PR #373, opened 2020-09-02
docs: fix macos install guide

foray1010:
`--with-lldb --with-toolchain` are removed from https://github.com/Homebrew/homebrew-core/blob/master/Formula/llvm.rb


RafaelGSS (approved):
LGTM.

PR #394, opened 2021-10-16
chore(ci): fix CI by disabling nightly builds

mmarchini:


mmarchini:
I'll get back to this later, I have no idea why so many tests are failing. I don't think it's environment related since it's kinda across the board, the test might just be flaky.

PR #415, opened 2022-09-24
src: fix warnings

kvakil:
Fix various warnings which show up when compiling with gcc. For the most part this doesn't change any semantics -- the only user-facing change is a bug fix in `Printer::Stringify` caused by a missing `return`.


No9:
LGTM - I'm rerunning the the flaky tests and then I'll land it.
I'll also review the other PRs and then look to do a release.

No9 (approved):

PR #404, opened 2022-08-14
changes to support Node.js v16 & v18

kvakil:
With this diff, our tests are much better. Previously they were entirely
failing on Node 16. Now:

* `stack-test` passes (12/12)
* `inspect-test` mostly passes (57/59)
* `frame-test` somewhat passes (7/25)

I marked the failing tests as skipped / optional.

There are a couple of big commits in this diff. It is best reviewed
commit-by-commit. They are in order:

1. Postmortem data fixes
   - `class_Map__constructor_or_backpointer__Object` is now
     `class_Map__constructor_or_back_pointer__Object` (note the extra
     `_`).
   - `class_Script__source__Object` is weirdly _not_ present in Node 16
     but _is_ present in both Node 14 and Node 18. I added a default to
     the correct value of 8.

2. `ScopeInfo` is no longer a `FixedArray` as of v8/v8@ecaac329. The
   `Length` field was also removed in v8/v8@f731e13f.
   - The length field removal shifted everything over by a slot, meaning
     that a bunch of offsets changed.
   - Since `ScopeInfo` is no longer a `FixedArray`, I changed callsites
     to use `HeapObject::LoadFieldValue` rather than `FixedArray::Get`.

3. Changes to V8 calling conventions
   - The arguments adapter frame no longer exists. Modified
     `frame-test.js` to match.
   - Arguments are no longer pushed "in reverse". Modified
     `JSFrame::LeaParamSlot` to match.

4. A minor fix to `nodejsVersion`, which was returning `[NaN]` when
   run on a release build.

5. Remove support for unboxed doubles, which fixes printing doubles
   since V8 broke our previous detection mechanism.

6. Remove a couple of tests which seemed unnecessary (`NativeModule`
   and `Promise`).

PR #416, opened 2022-09-24
src: remove unboxed double support completely

kvakil:
After #404, we no longer support unboxed doubles. Let's remove the relevant code.


No9 (approved):
LGTM - It's a very clean removal so landing

PR #417, opened 2022-09-24
remove more code used to support old Node.js versions

kvakil:
This removes ~100 lines of code which were used to support old versions of Node.js. Doing this revealed some more places where we are missing postmortem metadata.

This is best reviewed commit-by-commit.


No9 (approved):
LGTM - Everything seems straight forward. 
Going to ask for a second review just incase.

No9:
@RafaelGSS Would you mind giving this a review when you have a min. 
It looks good but just want a second opinion given the size of the change.

RafaelGSS (approved):
I don't have enough knowledge of this codebase to say if that will break any behaviour.  Once the CI passes, I'm good.

PR #418, opened 2022-09-24
src: fix PropertyAttributesEnum_DONT_DELETE typo

kvakil:
Frankly, this code is currently unused and it probably makes more sense to remove all of it entirely. But if we keep it, let's fix this typo I noticed.


No9 (approved):
Looks like a copy/paste error and the fix is straight forward.
LGTM

No9:
Thanks @kvakil - Simple fix so I'm merging

PR #419, opened 2022-09-24
test: skip flaky JSArrayBuffer test

kvakil:
This test failed a couple of times in #415, #417 and #418, so let's skip it for now.


No9 (approved):
LGTM - Happy that this is done as an optional rather than a comment out.

No9:
Merging and recorded here https://github.com/nodejs/llnode/issues/421

PR #420, opened 2022-09-24
test: blindly attempt to re-enable nightly

kvakil:
I disabled this in #404. This is a PR so I can futz around with it and try to figure out why the mirror isn't working.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -19,11 +19,10 @@ jobs:
           - version: 14.x
           - version: 16.x
           - version: 18.x
-          # These error with "Unexpected input(s) 'node-mirror' ...":
-          # - version: 19.x
-          #   mirror: https://nodejs.org/download/nightly
-          # - version: 19.x
-          #   mirror: https://nodejs.org/download/v8-canary
+          - version: 19.x
+            mirror: https://nodejs.org/download/nightly
+          - version: 19.x
+            mirror: https://nodejs.org/download/v8-canary
         # os: [ubuntu-latest, macos-latest]
         # Temporarily disable MacOS until
         # https://github.com/nodejs/node/issues/32981 is fixed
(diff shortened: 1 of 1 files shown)


kvakil:
It fails with this error:
```
Warning: Unexpected input(s) 'node-mirror', valid inputs are ['always-auth', 'node-version', 'architecture', 'check-latest', 'registry-url', 'scope', 'token', 'cache', 'cache-dependency-path', 'version']
Run No9/setup-node@mirror
Attempting to download 19.x...
Not found in manifest.  Falling back to download directly from Node
Error: Unable to find Node version '19.x' for platform linux and architecture x64.
```

(The warning probably isn't relevant, I think it just needs this patch:
```
diff --git a/action.yml b/action.yml
index e50fa0e..af619f2 100644
--- a/action.yml
+++ b/action.yml
@@ -29,6 +29,8 @@ inputs:
   version:
     description: 'Deprecated. Use node-version instead. Will not be supported after October 1, 2019'
     deprecationMessage: 'The version property will not be supported after October 1, 2019. Use node-version instead'
+  node-mirror:
+    description: 'Optional URL of an alternate Node.js mirror to use. Example: https://nodejs.org/download/nightly/'
 outputs:
   cache-hit: 
     description: 'A boolean value to indicate if a cache was hit'
```

but I don't think that it's actually causing the issue? https://github.com/actions/runner/issues/514 looks like the upstream issue.)

No9:
@kvakil This is failing because 19.x doesn't exist in https://nodejs.org/dist/ 
I'll take a look around the issues in core and update here with more info as I find it.

No9:
@kvakil So this was easy. 
There is no `19.x` because it's not due out until Oct 18th https://github.com/nodejs/node/issues/44650 
The issue states that `latest` releases are built against on `V8 10.2` same as `18.x` so there is not much point using `latest` in these tests. 
From the conversation it looks like there is a workaround to get tier1 support to do a release with `v8 10.5` so I would say there will be something in and around the 18th Oct date.
I've put a watch on that issue and we can re-run these tests when that lands.

No9:
Now that 19 is released i re-ran the tests 
The Ubuntu 18 are failing with:
```
node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by node)
Error: Process completed with exit code 1.
```
We should probably look at disabling Ubuntu 18 for the 19 tests?

PR #389, opened 2021-04-30
Use official `llvm-project` monorepo, expand LLDB versions check

trxcllnt:
* Fetch headers from the official [llvm-project](https://github.com/llvm/llvm-project) instead of the deprecated [llvm-mirror/lldb](https://github.com/llvm-mirror/lldb.git) repo.
* Check for LLDB versions `20.0 - 1.0.0` on Linux so we don't have to update LLDB versions for a few years.

Fixes https://github.com/nodejs/llnode/issues/350

PR #425, opened 2022-10-01
scripts: installation without sparse-checkout

kvakil:
Old versions of Ubuntu (and likely other LTS distros) have `git` versions which do not support `git-sparse-checkout`. `git-sparse-checkout` is used as of #389.

Dynamically check if `git sparse-checkout` is supported, and if not then fallback to cloning the whole repo. Cloning the whole repo is _slow_, but at least it works.

I tested this patch on Ubuntu 18.04 without any lldb headers installed, and it worked successfully.

### DIFF
--- a/scripts/lldb.js
+++ b/scripts/lldb.js
@@ -47,9 +47,17 @@ function getLibPath(libDir) {
 }
 
 /**
+ * Check if the installed version of git supports sparse checkout. Notably,
+ * Ubuntu 18.04 (which is not EOL at time of writing) has git 2.17, which does
+ * not support sparse-checkout.
+ */
+function doesGitSupportSparseCheckout() {
+  const returnCode = child_process.spawnSync('git', ['sparse-checkout', '--help']).status;
+  return returnCode === 0;
+}
+
+/**
  * Check out source code of the lldb for headers
- * TODO: The llvm project is probably moving to github soon at that point we
- * should stop using the mirror.
  * @param {string} lldbVersion Version of lldb, either like 3.9 or 39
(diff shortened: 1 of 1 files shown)


scripts/lldb.js
```diff
@@ -60,26 +68,39 @@ function cloneHeaders(lldbVersion, buildDir) {
...
+      // use `git clone --filter` in git v2.19 to only download `lldb` dir of `llvm-project` monorepo
+      // see: https://stackoverflow.com/a/52269934/3117331
+      // see: https://github.blog/2020-01-17-bring-your-monorepo-down-to-size-with-sparse-checkout/
+      child_process.execFileSync(
+          'git', ['clone',
+            '--depth', '1',
+            '--filter=blob:none',
+            '--sparse',
+            '--branch', lldbHeadersBranch,
+            'https://github.com/llvm/llvm-project.git',
```
tony-go, line 81: Maybe could we put that one into a variable ?

tony-go (approved):

scripts/lldb.js
No9: Hey @tony-go I'm looking to merge this.
Can you be a bit more specific with the point about the variable? 
I'm not 100% on what your ask is.

kvakil: I think the ask is to put the Github URL into a variable. I don't feel particularly strongly about it; feel free to do it before merging if you'd like.

No9: OK this has been around for a while with no updates. 
@tony-go I'm also not strongly pushed about changing the implementation.
I'll merge this and please feel free to PR your suggested change if you want.

Issue #422, opened 2022-09-28
bug(print-source): libc++abi: terminating with uncaught exception


tony-go:
**Describe the bug**
llnode crashes when I run `v8 i -s <addr>`.

**To reproduce**

- Having this file `crash.js`
```js
'use strict';

throw new Error('this is uncaught');
```
- Running `node --abort-on-uncaught-exception crash.js`
- Loaded core with llnode node/location -c core/location
- Ran command `v8 bt`:
```bash  
(llnode) v8 bt
 * thread #1: tid = 0x0000, 0x0000000101756ca5 node`v8::base::OS::Abort() (.cold.1) + 5
  * frame #0: 0x0000000101756ca5 node`v8::base::OS::Abort() (.cold.1) + 5
    frame #1: 0x000000010147fa77 node`v8::base::OS::Abort() + 23
    frame #2: 0x00000001006e10bb node`v8::internal::Isolate::CreateMessageOrAbort(v8::internal::Handle<v8::internal::Object>, v8::internal::MessageLocation*) + 171
    frame #3: 0x00000001006e0aae node`v8::internal::Isolate::ThrowInternal(v8::internal::Object, v8::internal::MessageLocation*) + 958
    frame #4: 0x0000000100ae0ecf node`v8::internal::Runtime_Throw(int, unsigned long*, v8::internal::Isolate*) + 47
    frame #5: 0x0000000100e8b319 <exit>
    frame #6: 0x0000000100f20d5a <stub>
    frame #7: 0x0000000100e1ddea (this=0x2f03a76913e9:<Object: Object>, 0x2f03a76913e9:<Object: Object>, 0x2f03a7693201:<function: require at node:internal/modules/cjs/helpers:1:10>, 0x2f03a76911e1:<Object: Module>, 0x2f03a768f159:<String: "/Users/jamiamer/...">, 0x2f03a7693131:<String: "/Users/jamiamer/...">) at /Users/jamiamer/projects/llnode-tuto/simple-crash.js:1:0 fn=0x00002f03a7692e11
...
```
- Ran command `v8 i -s 0x2f03a7693201`:
- Crash. 

**Expected behavior**
Having the source of the function printed in the console.

**Console output**
```bash
(llnode) v8 i -s 0x2f03a7693201
libc++abi: terminating with uncaught exception of type std::out_of_range: basic_string
PLEASE submit a bug report to https://developer.apple.com/bug-reporting/ and include the crash backtrace.
Stack dump:
[...]

PR #426, opened 2022-10-01
src: don't crash on invalid script positions

kvakil:
This "fixes" the segfault reported in #422, in the sense that you no longer get a segfault. However the printing does not actually work, i.e. you currently get an error like this:

```console
(lldb) v8 i -s 0x2196b1a09a29
error: Invalid source range, start_pos=3108, len=-3098, source_len=10
```

I'm deeming this better than segfaulting. We should really never be segfaulting as the coredump might be incomplete/partially corrupted. (Also, we already know function printing on v16 doesn't work right now.)

### DIFF
--- a/src/llv8.cc
+++ b/src/llv8.cc
@@ -362,8 +362,15 @@ std::string JSFunction::GetSource(Error& err) {
   }
   int64_t len = end_pos - start_pos;
 
-  std::string res = source_str.substr(start_pos, len);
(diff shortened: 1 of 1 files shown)


tony-go:
Thanks ^^

Definitely better than segfault. 😃

But we don't really know why that is happening. I'm correct?

No9:
LGTM to so merging

PR #431, opened 2023-09-30
Add "v8 version" subcommand to llnode

### LINKED ISSUE #194: "v8 version" or "llnode --version"
Today there's no easy way to get llnode's version, making it harder to report issues with detailed information. We could either have a llnode command `v8 version` or a command line argument `--version` reporting llnode's version, maybe showing LLDB's version as well.

vprab:
Fixes: https://github.com/nodejs/llnode/issues/194

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -62,6 +62,7 @@
       "src/node-constants.cc",
       "src/settings.cc",
     ],
+    "cflags": [ '-DLLNODE_VERSION=<!(grep -Po \'"version": \K(".*")\' <(module_root_dir)/package.json)' ],
     "conditions": [
       [ "OS == 'win'", {
         "sources": [
--- a/src/llnode.cc
+++ b/src/llnode.cc
@@ -394,6 +394,12 @@ std::string GetActiveRequestsCmd::GetResultMessage(node::Environment* env,
   return result_message.str();
 }
 
+bool GetLLNodeVersionCmd::DoExecute(SBDebugger d, char** cmd,
+				    SBCommandReturnObject& result) {
+  result.Printf("llnode version %s", LLNODE_VERSION);
+  return true;
+}
+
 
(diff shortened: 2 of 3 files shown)


binding.gyp
```diff
@@ -62,6 +62,7 @@
       "src/node-constants.cc",
       "src/settings.cc",
     ],
+    "cflags": [ '-DLLNODE_VERSION=<!(grep -Po \'"version": \K(".*")\' <(module_root_dir)/package.json)' ],
```
zhxinyu, line 65: Will the new feature support in Mac OS system? As far as I know, option P is not supported in `grep` in Mac OS, neither as `\K`

zhxinyu (changes requested):

binding.gyp
vprab: This is a compile-time step, so technically the feature would support any OS once we have a built artifact, but you're right that this wouldn't build on MacOS. We can change this line to:
```"cflags": [ '-DLLNODE_VERSION=<!(grep \\"version\\" <(module_root_dir)/package.json | sed \'s/.*\\("[0-9\\.]*"\\).*/\\1/g\')' ],```
if that works on Mac? Although I guess that also brings up the issue of Windows...

vprab:
Tagging @danielleadams for visibility 😄

binding.gyp
zhxinyu: Yes. I confirm that it should work on Mac. and  I think  the double `\` in `\\"version\\"` can be replaced by single `\`. 

In particular, I tried `grep "version" < package.json | sed 's/.*\("[0-9\.]*"\).*/\1/g'` and `echo "$(grep \"version\" < package.json | sed 's/.*\("[0-9\.]*"\).*/\1/g')"`in cli.

PR #433, opened 2024-03-07
Support llvm dev releases

trxcllnt:
This is a followup to https://github.com/nodejs/llnode/pull/389:

* Updates the range of LLVM versions we check (from 20, which is next dev version, to 99)
* Support downloading headers if the detected `lldb` version is the current dev release
    * i.e. doesn't have a `release/XY.x` branch, but does have a `llvmorg-XY-init` tag
* Update the PR action to test ubuntu-22.04 + LLVM versions up to current dev (v19)

If we wanted to be fancy, we could populate the supported versions and CI matrix by parsing `git ls-remote --tags https://github.com/llvm/llvm-project.git 'refs/tags/llvmorg-*-init'`.

Let me know if that's something you'd like me to add!

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -151,7 +151,7 @@ jobs:
       - name: npm install, build, and test
         run: |
           sudo apt-get -qq update
-          sudo apt-get install lldb-6.0 liblldb-6.0-dev lcov -y
-          sudo ln -s $(which lldb-6.0) /usr/bin/lldb
+          sudo apt-get install lldb-11 liblldb-11-dev lcov -y
+          sudo ln -s $(which lldb-11) /usr/bin/lldb
           npm install
           npm run linter
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -112,9 +112,10 @@ jobs:
 
           if [[ -n "$(which lldb-${{ matrix.llvm }})" ]]; then
             sudo ln -s "$(which lldb-${{ matrix.llvm }})" /usr/bin/lldb
+            python_version="$(python3 --version 2>&1 | cut -d' ' -f2 | cut -d'.' -f3 --complement)";
             # Is `/usr/lib/lib` here correct?
-            sudo mkdir -p /usr/lib/lib/python3.8
-            sudo ln -s /usr/lib/llvm-${{ matrix.llvm }}/lib/python3.8/site-packages /usr/lib/lib/python3.8/site-packages
+            sudo mkdir -p /usr/lib/lib/python${python_version}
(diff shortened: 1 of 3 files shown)


.github/workflows/push.yml
```diff
@@ -74,8 +112,10 @@ jobs:
 
           if [[ -n "$(which lldb-${{ matrix.llvm }})" ]]; then
             sudo ln -s "$(which lldb-${{ matrix.llvm }})" /usr/bin/lldb
-            sudo mkdir -p /usr/lib/lib/python3.8
-            sudo ln -s /usr/lib/llvm-${{ matrix.llvm }}/lib/python3.8/site-packages /usr/lib/lib/python3.8/site-packages
+            python_version="$(python3 --version 2>&1 | cut -d' ' -f2 | cut -d'.' -f3 --complement)";
+            # Is `/usr/lib/lib` here correct?
+            sudo mkdir -p /usr/lib/lib/python${python_version}
+            sudo ln -s /usr/lib/llvm-${{ matrix.llvm }}/lib/python${python_version}/site-packages /usr/lib/lib/python${python_version}/site-packages
```
trxcllnt, line 113: Left a comment here, not sure what symlinking to `/usr/lib/lib` is doing?

.github/workflows/push.yml
```diff
@@ -112,7 +152,7 @@ jobs:
       - name: npm install, build, and test
         run: |
           sudo apt-get -qq update
-          sudo apt-get install lldb-6.0 liblldb-6.0-dev lcov -y
-          sudo ln -s $(which lldb-6.0) /usr/bin/lldb
+          sudo apt-get install lldb-11 liblldb-11-dev lcov -y
+          sudo ln -s $(which lldb-11) /usr/bin/lldb
```
trxcllnt, line 154: Updated these to `lldb-11` because ubuntu-latest is now jammy

.github/workflows/push.yml
```diff
@@ -44,25 +48,59 @@ jobs:
...
+              case "${{ matrix.llvm }}" in
+                11) use_llvm_repos=0;;
+                12) use_llvm_repos=0;;
+                13) use_llvm_repos=0;;
+                14) use_llvm_repos=0;;
+                15) use_llvm_repos=0;;
+                *) use_llvm_repos=1;;
+              esac
+              ;;
+            *) use_llvm_repos=1;;
```
trxcllnt: This defaults to using the LLVM repos _unless_ it's one of the versions packaged by Canonical. This should be a limited set, compared to the previous way which would need exceptions for every version _not_ packaged by Canonical.

trxcllnt:
It appears the lldb headers in llvm-19 now include headers generated by the build, so downloading them from github is insufficient to build llnode. I'm not sure what to do about that.

@No9 do you have any thoughts on this PR?

